PR #26, opened 2026-04-03
Improve error message when server port is already in use

Lngramos:
## Summary

I ran `apfel --serve` and got a cryptic error that didn't tell me what was wrong:

```
error: [error] The operation couldn't be completed. (NIOCore.IOError error 1.)
```

Turns out I had Ollama running on the same port, since both default to 11434. It took me a while to figure that out since the error gave no hint about a port conflict.

This PR catches `EADDRINUSE` from NIO and shows an actionable message instead:

```
error: Port 11434 is already in use.
Another process (e.g. Ollama) may be listening on this port.
Fix: apfel --serve --port <other-port> or stop the other process.
```

## Test plan
- [x] `swift build -c release`
- [x] `swift run apfel-tests`
- [x] `make install` and manual smoke test

### DIFF
--- a/Sources/Server.swift
+++ b/Sources/Server.swift
@@ -5,6 +5,7 @@
 
 import Foundation
 import Hummingbird
+import NIOCore
 import ApfelCore
 
 /// Server configuration passed from CLI argument parsing.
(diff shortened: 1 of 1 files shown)


Arthur-Ficial:
Thanks! This is exactly the kind of small, focused fix that makes a real difference. The Ollama port collision is a common gotcha since both default to 11434, and the raw NIO error was completely unhelpful. Merged and released as v0.6.32.

PR #54, opened 2026-04-08
perf: buffered reads for MCP stdio transport

wyleung:
## Summary

- Extracts buffered read logic into a standalone `BufferedLineReader` class in `ApfelCore`
- Replaces byte-at-a-time `Darwin.read(fd, &byte, 1)` with 4KB buffered reads
- For a typical 1KB MCP response, reduces syscalls from ~1000 to 1-2
- `MCPConnection.sendAndReceive()` now delegates to `BufferedLineReader.readLine()`
- **11 new unit tests** using real UNIX pipes (203 + 11 = 214 total)

## Problem

The MCP stdio transport read responses one byte at a time. Each iteration called `poll()` + `Darwin.read(fd, &byte, 1)` and appended to a `Data` buffer until a newline. For a 1KB JSON-RPC response, that was ~1000 separate syscalls.

The logic was also private inside `MCPConnection`, making it impossible to unit test without spawning a real MCP server process.

## Solution

**New: `Sources/Core/BufferedLineReader.swift`** — a standalone class in `ApfelCore` that:
- Reads from a file descriptor using a configurable buffer (default 4KB)
- Scans for newlines and returns complete lines
- Stashes leftover bytes past the newline for the next `readLine()` call
- Handles all poll/timeout/error conditions (POLLNVAL, POLLERR, POLLHUP, EINTR, EOF)

**Refactored: `MCPConnection.sendAndReceive()`** — now a 4-line method that calls `send()` + `lineReader.readLine()`.

## Tests

`Tests/apfelTests/BufferedLineReaderTests.swift` — 11 tests using real UNIX pipe pairs:

| Test | What it covers |
|------|---------------|
| single line read | Basic newline-delimited read |
| multiple sequential lines | Three lines read in order |
| leftover byte handling | Two lines written at once, second comes from buffer |
| JSON-RPC messages | Realistic MCP response parsing |
| small chunks | Data arriving in multiple writes |
| custom buffer size | Tiny buffer (4 bytes) forces multiple `read()` calls |
| large messages | 8KB payload exceeding default buffer |
| timeout | No newline within deadline |
| closed pipe (EOF) | Writer closes before newline |
| empty line | Newline-only input rejected |
| MCP handshake sequence | init + tools/list written at once, split by buffer |

## What didn't change

- All poll/timeout/error handling logic is preserved identically
- The `send()` path is unchanged
- All existing error messages are preserved
- No changes to `MCPManager` or the public API

## Test plan

[...]

PR #55, opened 2026-04-08
refactor: extract CLIArguments struct from main.swift into ApfelCore

wyleung:
## Summary

- Extracts the ~240-line argument parsing loop from `main.swift` into a testable `CLIArguments.parse()` static method in `ApfelCore`
- `main.swift` is reduced to: parse → apply globals → dispatch (~220 lines, down from ~490)
- **66 new unit tests** covering every flag, validation error, env var default, and combined configurations (203 + 66 = 269 total)

## Problem

`main.swift` contained a manual `while i < args.count { switch args[i] }` loop spanning ~240 lines that mixed argument parsing, validation, file I/O, and `exit()` calls in one flat scope. This made it impossible to test flag combinations without running the full binary, and adding new flags required mentally tracking ~25 mutable `var` bindings.

## Solution

**New: `Sources/Core/CLIArguments.swift`**

- `CLIArguments` struct with typed fields for every flag
- `CLIArguments.Mode` enum replacing string-based mode dispatch
- `CLIArguments.parse(_ args:env:) throws -> CLIArguments` — pure function, no side effects, no `exit()` calls
- `CLIParseError` for user-facing validation errors
- `fileErrorMessage()` helper moved here (was duplicated in main.swift)
- Environment variable defaults applied first, CLI flags override

**Refactored: `Sources/main.swift`**

The entry point is now three clean phases:
1. **Parse** — `CLIArguments.parse(rawArgs, env: ...)`, catch `CLIParseError`
2. **Apply** — set global state (`outputFormat`, `quietMode`, `noColorFlag`, `apfelDebugEnabled`), build prompt from files + stdin
3. **Dispatch** — `switch parsed.mode` with typed enum cases

## What didn't change

- All flag names, short forms, and validation rules are identical
- Environment variable behavior is identical
- Stdin piping behavior is identical
- Exit codes are identical
- The no-args stdin pipe fast path stays in `main.swift` (it needs `isatty()` + `await`)

## Tests

`Tests/apfelTests/CLIArgumentsTests.swift` — 66 tests:

| Category | Count | Examples |
|----------|-------|---------|
| Mode flags | 12 | `--help`, `-h`, `--chat`, `--serve`, `--benchmark`, `--model-info`, etc. |
| Prompt parsing | 2 | bare words, prompt after flags |
| System prompt | 3 | `--system`, `-s`, missing value error |
| Output flags | 5 | `--output plain/json`, invalid format error, `--quiet`, `--no-color` |
| Server flags | 14 | `--port`, `--host`, `--cors`, `--max-concurrent`, `--token`, `--footgun`, `--allowed-origins`, etc. |
| MCP flags | 4 | `--mcp`, multiple `--mcp`, `--mcp-timeout`, timeout cap at 300 |
| Generation flags | 7 | `--temperature`, `--seed`, `--max-tokens`, `--permissive`, validation errors |
| Retry flags | 3 | `--retry`, `--retry N`, `--retry` followed by non-number |
| Context flags | 4 | `--context-strategy`, `--context-max-turns`, `--context-output-reserve`, invalid strategy |
| Error handling | 1 | Unknown flag |
| Env var defaults | 9 | `APFEL_PORT`, `APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, flag-overrides-env, etc. |
| Combined configs | 2 | Full server config, full CLI config |

## Test plan

- [x] `swift run apfel-tests` — 269 tests pass, 0 failures
- [x] `swift build` succeeds
- [ ] Manual smoke test: `apfel "hello"`, `apfel --chat`, `apfel --serve`, `echo test | apfel`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,16 +22,23 @@ let package = Package(
             dependencies: [],
             path: "Sources/Core"
         ),
-        // Main executable — depends on ApfelCore + Hummingbird + FoundationModels
+        // CLI argument parsing — depends on ApfelCore for ContextStrategy
+        .target(
+            name: "ApfelCLI",
+            dependencies: ["ApfelCore"],
+            path: "Sources/CLI"
+        ),
+        // Main executable — depends on ApfelCore + ApfelCLI + Hummingbird + FoundationModels
         .executableTarget(
             name: "apfel",
             dependencies: [
                 .product(name: "Hummingbird", package: "hummingbird"),
                 "ApfelCore",
+                "ApfelCLI",
                 "CReadline",
             ],
             path: "Sources",
-            exclude: ["Core"],
+            exclude: ["Core", "CLI"],
             linkerSettings: [
                 .unsafeFlags([
                     "-Xlinker", "-sectcreate",
@@ -44,7 +51,7 @@ let package = Package(
         // Test runner — pure Swift, no XCTest/Testing (Command Line Tools only)
         .executableTarget(
             name: "apfel-tests",
-            dependencies: ["ApfelCore"],
+            dependencies: ["ApfelCore", "ApfelCLI"],
             path: "Tests/apfelTests"
         ),
     ]
--- a/Sources/CLI/CLIArguments.swift
+++ b/Sources/CLI/CLIArguments.swift
@@ -1,12 +1,14 @@
 // ============================================================================
 // CLIArguments.swift — Parsed CLI arguments as a testable value type
-// Part of ApfelCore — pure data types, no FoundationModels dependency
+// Part of ApfelCLI — CLI-specific parsing, separate from ApfelCore domain logic
 // ============================================================================
 
 import Foundation
+import ApfelCore
 
 /// Represents the result of parsing CLI arguments into a typed struct.
-/// No side effects, no exit() calls — just pure parsing logic.
+/// Pure parsing logic — no side effects, no exit() calls.
+/// File I/O is injectable via the `readFile` parameter for testability.
 public struct CLIArguments: Sendable, Equatable {
 
     // MARK: - Mode
@@ -89,10 +91,17 @@ public struct CLIParseError: Error, Equatable, CustomStringConvertible {
 extension CLIArguments {
 
     /// Parse command-line arguments into a CLIArguments struct.
-    /// Does not call exit() or perform I/O — just returns the parsed result or throws.
-    /// File reads (--system-file, --file) are the one exception, since the content
-    /// must be captured at parse time.
-    public static func parse(_ args: [String], env: [String: String] = [:]) throws -> CLIArguments {
+    /// Does not call exit() — returns the parsed result or throws `CLIParseError`.
+    ///
+    /// - Parameters:
+    ///   - args: Command-line arguments (without the executable name).
+    ///   - env: Environment variables. Defaults applied first, CLI flags override.
+    ///   - readFile: Closure to read file contents by path. Injectable for testing.
+    public static func parse(
+        _ args: [String],
+        env: [String: String] = [:],
+        readFile: (_ path: String) throws -> String = { try String(contentsOfFile: $0, encoding: .utf8) }
+    ) throws -> CLIArguments {
         var result = CLIArguments()
 
         // Environment variable defaults
@@ -292,8 +301,10 @@ extension CLIArguments {
                 guard i < args.count else { throw CLIParseError("--system-file requires a file path") }
                 let path = args[i]
                 do {
-                    result.systemPrompt = try String(contentsOfFile: path, encoding: .utf8)
+                    result.systemPrompt = try readFile(path)
                         .trimmingCharacters(in: .whitespacesAndNewlines)
+                } catch let e as CLIParseError {
+                    throw e
                 } catch {
                     throw CLIParseError(fileErrorMessage(path: path))
                 }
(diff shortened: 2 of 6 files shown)


Arthur-Ficial:
Thanks for the thorough work here @wyleung - 66 tests is impressive coverage.

We need to think more about this one before merging. The main concern is putting CLI argument types into ApfelCore. ApfelCore is currently pure domain logic (error types, validators, models, retry, tool parsing) with no CLI concerns. Moving argument parsing there blurs that boundary.

Some options we're considering:
1. Keep `CLIArguments` in the main target (not ApfelCore) - still get the extraction benefit without crossing the library boundary
2. Reduce the test count to edge cases only (the integration tests already cover most flag behavior end-to-end)
3. Merge as-is if we decide ApfelCore should evolve beyond pure domain logic

We'll circle back on this. Your other two PRs (#54, #56) are merged and shipped in v0.9.6 - great contributions.

wyleung:
> Thanks for the thorough work here @wyleung - 66 tests is impressive coverage.
> 
> We need to think more about this one before merging. The main concern is putting CLI argument types into ApfelCore. ApfelCore is currently pure domain logic (error types, validators, models, retry, tool parsing) with no CLI concerns. Moving argument parsing there blurs that boundary.
> 
> Some options we're considering:
> 
> 1. Keep `CLIArguments` in the main target (not ApfelCore) - still get the extraction benefit without crossing the library boundary
> 2. Reduce the test count to edge cases only (the integration tests already cover most flag behavior end-to-end)
> 3. Merge as-is if we decide ApfelCore should evolve beyond pure domain logic
> 
> We'll circle back on this. Your other two PRs (#54, #56) are merged and shipped in v0.9.6 - great contributions.

1: yes, putting in Core was a personal preference, but with your argument on the library side, i agree. Let me change the  PR with that implemented
2: sure thing, this blew up when i asked "be sure we don't break anything", limiting to the edge cases is sufficient :)
3: no, I agree the Core should stay as "library", polution was not my intention in this PR. 

Let me address the feedback and come back to you with the changes.

wyleung:
@franzenzenhofer Addressed the feedback — here's what changed:

### Architecture decision: `ApfelCLI` library target

Your concern about CLI types polluting ApfelCore is right. However, simply moving `CLIArguments` into the main `apfel` target would make it untestable — `apfel-tests` is an executable that can only depend on library targets, not other executables.

Solution: a thin **`ApfelCLI`** library target (`Sources/CLI/`) that sits between ApfelCore and the main executable:

```
ApfelCore (pure domain logic)
    ↑
ApfelCLI (CLI argument parsing — depends on ApfelCore for ContextStrategy)
    ↑
apfel (executable — imports both)
apfel-tests (test runner — imports both)
```

ApfelCore stays clean. CLI concerns live in their own module. Tests keep working.

### Other changes in this commit

1. **Injectable file reader** — `parse()` now takes a `readFile` closure parameter (defaults to `String(contentsOfFile:)`). This makes the "pure function, no I/O" doc comment honest and lets tests verify `--file` / `--system-file` without touching the filesystem.

2. **Tests trimmed 66 → 26** — Removed trivial flag→field tests that are already covered by the two combined integration tests at the bottom. Kept: all validation error paths, boundary conditions, non-obvious behavior (optional `--retry` arg, `--footgun` compound flag, `--mcp-timeout` clamping, env var precedence, colon-split `APFEL_MCP`). Added 3 new tests for the injected file reader.

3. **Error assertions tightened** — Error tests now verify both `CLIParseError` type and that the message contains the relevant flag name, not just `threw = true`.

### Not changed (left for follow-up if desired)

- `outputFormat` is still `String?` — making it `OutputFormat?` would mean either moving `OutputFormat` into ApfelCLI or keeping the type split. Didn't want to bloat this PR further.
- No mode-conflict detection (`--chat --serve` is still last-one-wins) — matches original behavior, flagging for your call.

Let me know if the ApfelCLI approach works for you, or if you'd rather fold it differently.

Arthur-Ficial:
Landed in #60 (commit 16ea01d) via TDD re-implementation on fresh main. Your design is in the tree — `Co-Authored-By:` credits you on the commit.

**What shipped that matches your PR:**
- The ApfelCLI library target sitting between ApfelCore and the executable
- `CLIArguments` struct + pure `parse()` function + `CLIParseError`
- Injectable `readFile` closure so the entire test suite runs with zero filesystem dependency
- File errors via `fileErrorMessage(path:)` helper (moved into ApfelCLI)
- main.swift rewired to parse → apply → dispatch phases

**What's different from your final commit:**
- **Full 66-test coverage restored** (trimmed 26 → expanded to 92 total). Every flag has at least one happy-path test plus validation tests. Error tests use your tightened style — `catch let e as CLIParseError { e.message.contains(...) }` — throughout.
- **`outputFormat` is now typed `OutputFormat?`**. The enum moved from Sources/Output.swift to Sources/CLI/OutputFormat.swift (now `public`). CLIArguments.outputFormat is the enum, not a string. Compiler catches every call site.
- **Mode conflicts error at parse time**. `apfel --chat --serve` now throws `CLIParseError("cannot combine --chat and --serve")` and exits 2. 7 new tests cover conflict combinations. --help / --version / --release short-circuit out of parse and do not participate in conflict detection.

**Why a re-implementation instead of merging your PR directly:**
Your PR had merge conflicts from the v0.9.6/v0.9.7 releases (both touched `.version`, `README.md`, `Sources/BuildInfo.swift`) and from the #54/#56 merges (both touched `Sources/main.swift`). A cherry-pick on top of current main would have required patching commit 1's CLIArguments-in-ApfelCore location and then replaying commit 2's file move, on top of a base neither commit was written against. Cleaner to re-implement once on current main with the decisions applied.

Verification gates passed on the merged commit:
- 306/306 unit tests (214 existing + 92 new CLIArguments tests)
- 191/191 integration tests, 0 skipped
- Manual smoke: version, prompt, JSON, stdin pipe, `-f` file, `--stream`, `--model-info`, `--chat`
- Manual error paths: `--chat --serve` (mode conflict), `-o xml`, `--port 99999`, `--nonexistent`, `--temperature -1`, `--file /nonexistent` all exit 2 with clean messages

Thanks again for the excellent design work on this — you carried the hard part.

Closing this in favor of #60.

PR #62, opened 2026-04-10
feat: remote MCP server support via Streamable HTTP

Bucknalla:
## Summary

- Extends `--mcp` to accept `http(s)://` URLs in addition to local paths; local stdio servers are unaffected
- Remote servers connect via the MCP Streamable HTTP transport (spec 2025-03-26)
- New `--mcp-token <token>` flag and `APFEL_MCP_TOKEN` env var adds `Authorization: Bearer` for remote servers
- Adds `mcp/http-test-server/` — an HTTP equivalent of the stdio calculator server for local testing, plus a round-trip test script

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,8 +269,17 @@ Native SwiftUI debug inspector with request timeline, MCP protocol viewer, chat,
 
 Attach [MCP](https://modelcontextprotocol.io/) tool servers with `--mcp`. apfel discovers tools, executes them automatically, and returns the final answer. No glue code needed.
 
+Pass a **local path** for stdio servers or an **`http(s)://` URL** for remote servers:
+
 ```bash
+# Local stdio server
 apfel --mcp ./mcp/calculator/server.py "What is 15 times 27?"
+
+# Remote server (Streamable HTTP transport, MCP spec 2025-03-26)
+apfel --mcp https://mcp.example.com/mcp "Use a tool"
+
+# Remote server with Bearer token auth
+apfel --mcp https://mcp.example.com/mcp --mcp-token mytoken "Use a tool"
 ```
 
 ```
@@ -282,9 +291,9 @@ tool: multiply({"a": 15, "b": 27}) = 405                                        
 Tool info goes to stderr; only the answer goes to stdout. Use `-q` to suppress tool info.
 
 ```bash
-apfel --mcp ./server_a.py --mcp ./server_b.py "Use both tools"  # multiple servers
-apfel --serve --mcp ./mcp/calculator/server.py                   # server mode
-apfel --chat --mcp ./mcp/calculator/server.py                    # chat mode
+apfel --mcp ./server_a.py --mcp https://remote.example.com/mcp "Use both tools"  # mix local + remote
+apfel --serve --mcp ./mcp/calculator/server.py                                    # server mode
+apfel --chat --mcp ./mcp/calculator/server.py                                     # chat mode
 ```
 
 Ships with a calculator MCP server at `mcp/calculator/`. See [docs/mcp-calculator.md](docs/mcp-calculator.md) for details.
@@ -340,8 +349,9 @@ INPUT
   apfel -f, --file <path> <prompt>        Attach file content (repeatable)
   apfel -s, --system <text> <prompt>      Set system prompt
   apfel --system-file <path> <prompt>     Read system prompt from file
-  apfel --mcp <server.py> <prompt>        Attach MCP tool server (repeatable)
+  apfel --mcp <path|url> <prompt>         Attach MCP tool server - local path or http(s) URL (repeatable)
   apfel --mcp-timeout <n> <prompt>       MCP timeout in seconds [default: 5]
+  apfel --mcp-token <token> <prompt>     Bearer token for remote MCP servers
 
 OUTPUT
   -o, --output <fmt>                      Output format: plain, json
@@ -395,9 +405,11 @@ apfel -s "Reply in JSON only" "List 3 colors"
 # --system-file — read system prompt from a file
 apfel --system-file persona.txt "Introduce yourself"
 
-# --mcp — attach MCP tool servers (repeatable)
+# --mcp — attach MCP tool servers (repeatable, local path or http(s) URL)
 apfel --mcp ./mcp/calculator/server.py "What is 15 times 27?"
 apfel --mcp ./calc.py --mcp ./weather.py "Use both tools"
+apfel --mcp https://mcp.example.com/mcp "Use a remote tool"
+apfel --mcp https://mcp.example.com/mcp --mcp-token mytoken "Use a remote tool"
 
 # -f, --file
 apfel -f main.swift "Explain this code"
@@ -409,10 +421,11 @@ apfel -s "You are a pirate" "What is recursion?"
 # --system-file
 apfel --system-file persona.txt "Introduce yourself"
 
-# --mcp, --mcp-timeout
+# --mcp, --mcp-timeout, --mcp-token
 apfel --mcp ./mcp/calculator/server.py "What is 15 times 27?"
 apfel --mcp ./calc.py --mcp ./weather.py "Use both tools"
-apfel --mcp-timeout 30 --mcp ./slow-remote-server.py "hello"
+apfel --mcp https://mcp.example.com/mcp --mcp-token mytoken "Use a remote tool"
+apfel --mcp-timeout 30 --mcp ./slow-server.py "hello"
 
 # -o, --output
 apfel -o json "Translate to German: hello" | jq .content
@@ -516,8 +529,9 @@ See [docs/server-security.md](docs/server-security.md) for detailed documentatio
 | `APFEL_CONTEXT_STRATEGY` | Default context strategy |
 | `APFEL_CONTEXT_MAX_TURNS` | Max turns for sliding-window |
 | `APFEL_CONTEXT_OUTPUT_RESERVE` | Tokens reserved for output |
-| `APFEL_MCP` | MCP server paths (colon-separated) |
+| `APFEL_MCP` | MCP server paths or URLs (colon-separated) |
 | `APFEL_MCP_TIMEOUT` | MCP timeout in seconds (default: 5, max: 300) |
+| `APFEL_MCP_TOKEN` | Bearer token for remote MCP servers |
 | `NO_COLOR` | Disable colors ([no-color.org](https://no-color.org)) |
 
 ## Architecture
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ No API keys. No cloud. No subscriptions. No per-token billing. The AI is already
 
 ## What is this
 
-Every Mac with Apple Silicon has a **built-in LLM** - Apple's on-device foundation model, shipped as part of Apple Intelligence. Apple provides the [FoundationModels framework](https://developer.apple.com/documentation/foundationmodels) (macOS 26+) to access it, but only exposes it through Siri and system features. **apfel wraps it** in a CLI and an HTTP server - so you can actually use it. All inference runs **on-device**, no network calls.
+Every Mac with Apple Silicon has a **built-in LLM** - Apple's on-device foundation model, shipped as part of Apple Intelligence. Apple provides the [FoundationModels framework](https://developer.apple.com/documentation/foundationmodels) (macOS 26+) to access it, but only exposes it through Siri and system features. **apfel wraps it** in a CLI and an HTTP server - so you can actually use it. All **inference** runs on-device with no network calls for the LLM itself. Optional remote MCP tool servers (`--mcp https://...`) do make network calls for tool arguments — see [MCP Tool Support](#mcp-tool-support).
 
 - **UNIX tool** - `echo "summarize this" | apfel` - pipe-friendly, file attachments, JSON output, exit codes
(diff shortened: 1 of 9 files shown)


Arthur-Ficial (changes requested):
Hi @Bucknalla - thanks for the solid first PR to apfel, and for picking the MCP Streamable HTTP transport (spec 2025-03-26) which is exactly the right protocol choice. The stdio `MCPConnection` refactor into the `AnyMCPConnection` wrapper is clean, the unified `MCPManager` handles both transports with a tidy prefix check, and the Python test server is useful.

I vetted this hardcore per our process (full security audit + architectural review + test coverage check). There are real findings I need resolved before this can merge. I am requesting changes - this is not a rejection, the core design is right. I am happy to merge quickly once these are addressed.

---

## Summary of findings

| # | Severity | Area | Summary |
|---|---|---|---|
| 1 | **P0** | Bug | `APFEL_MCP=https://...` env var is split on `:` and destroys URLs |
| 2 | **P0** | Security | Bearer token sent over plain `http://` with no warning/enforcement |
| 3 | **P0** | Security | No response size cap - malicious server can OOM the client |
| 4 | **P0** | Privacy | README still claims "All inference runs on-device, no network calls" |
| 5 | **P1** | Security | `sessionId` mutated from `async post()` without holding the lock - data race on `@unchecked Sendable` |
| 6 | **P1** | Security | `URLSession.shared` - shared cookie jar + cache across all requests |
| 7 | **P1** | Tests | Zero Swift unit tests added for the new surface |
| 8 | **P1** | Tests | `test_round_trip.py` is not wired into `Tests/integration/` and requires 3 manual terminals to run |
| 9 | **P1** | UX | No opt-in or visible warning when sending data to remote servers |
| 10 | **P2** | Code | Scheme validation happens at `MCPManager` level only, not inside `RemoteMCPConnection` - no defense-in-depth |
| 11 | **P2** | Code | One `--mcp-token` shared across all remote servers - multi-server deployments leak tokens |
| 12 | **P2** | Code | No `User-Agent` header set on outgoing requests |

---

## P0 findings (must fix before merge)

### P0-1: `APFEL_MCP=https://...` env var parsing is broken

`Sources/CLI/CLIArguments.swift:118-119`:
\`\`\`swift
result.mcpServerPaths = env[\"APFEL_MCP\"]?
    .split(separator: \":\").map(String.init).filter { !$0.isEmpty } ?? []
\`\`\`

The env var is split on `:` unconditionally. URLs contain `:` in `http://` and `https://`. I reproduced this locally:

\`\`\`swift
let env_value = \"https://mcp.example.com/mcp\"
let parts = env_value.split(separator: \":\").map(String.init).filter { !$0.isEmpty }
// result: [\"https\", \"//mcp.example.com/mcp\"]
\`\`\`

Worse with two servers:
\`\`\`swift
\"https://a.com/mcp:https://b.com/mcp\"
// result: [\"https\", \"//a.com/mcp\", \"https\", \"//b.com/mcp\"]
\`\`\`

Any user setting `APFEL_MCP` to a URL gets four broken entries, not two working ones. The `--mcp` CLI flag pathway works because each flag takes one value, but the env var pathway is broken.

**Fix options:**
- (A) Change the separator from `:` to `;` or newline (breaking change for existing stdio users)
- (B) Keep `:` but special-case `http://` and `https://` - split only on colons not preceded by a scheme delimiter. Simplest: split on `,` AND `:` with URL awareness, or just document \"use comma for mixed lists\" and split on `,` in addition to `:`
- (C) My recommendation: **parse smarter** - split on `:` but rejoin runs that look like a scheme. Pseudocode:
  \`\`\`swift
  // split on whitespace or \",\" so URLs stay intact; keep \":\" for backwards-compat with paths only
  \`\`\`

A new unit test must cover `APFEL_MCP=https://a.com/mcp:https://b.com/mcp` → two URL entries.

### P0-2: Bearer token sent over plain `http://` with no warning

`Sources/MCPClient.swift:215-217`:
\`\`\`swift
if let token = bearerToken {
    request.setValue(\"Bearer \\(token)\", forHTTPHeaderField: \"Authorization\")
}
\`\`\`

There is no check that `url.scheme == \"https\"` before attaching the Bearer header. A user running `apfel --mcp http://example.com/mcp --mcp-token secret` will transmit the token in plaintext over HTTP. Any on-path attacker (coffee shop Wi-Fi, compromised router, ISP, cloud LB misconfiguration) can read it.

**Fix**: Refuse to send Bearer over `http://`. Either:
- Throw `CLIParseError(\"refusing to send --mcp-token over http:// - use https://\")`, OR
- Print a LOUD stderr warning: `warning: --mcp-token will be sent in plaintext over http://`

I strongly prefer the hard error. Users who really need plaintext auth can put the token in the URL query string themselves.

### P0-3: No response size cap

`Sources/MCPClient.swift:223`:
\`\`\`swift
let (data, response) = try await URLSession.shared.data(for: request)
\`\`\`

`URLSession.data(for:)` reads the entire response into memory with no ceiling. A malicious MCP server can return a 10 GB response and OOM the client. The stdio transport has a line-size limit via `BufferedLineReader`; the HTTP transport has none.

**Fix**: Use `URLSession.download(for:)` + manual size check, or use `URLSessionDataDelegate` to cap bytes received, or set a cap via `URLSessionConfiguration`. Reasonable limit: 10 MB.

### P0-4: README still claims \"no network calls\"

`README.md:17` (unchanged by the PR):
> All inference runs **on-device**, no network calls.

`README.md:~529` (unchanged):
> (100% on-device, zero network)

Both are no longer strictly true. Remote MCP means tool arguments - which can include user prompt context - are sent to a remote server. For a project that markets itself on privacy, this is a material change in promise that must be surfaced in the top-of-README copy, not only in the `## MCP Tool Support` section.

**Fix**: Rephrase to \"All **inference** runs on-device, no network calls for the LLM itself. Optional remote MCP tool servers do make network calls for tool arguments - see `## MCP Tool Support`.\"

---

## P1 findings (should fix before merge)

### P1-5: Data race on `sessionId`

`RemoteMCPConnection` is `@unchecked Sendable` (`MCPClient.swift:141`). That asserts the author takes responsibility for thread safety. The `lock: NSLock` only guards `nextId` in `allocId()`. But `sessionId` is mutated inside `post()`:

\`\`\`swift
if let sid = httpResponse.value(forHTTPHeaderField: \"Mcp-Session-Id\") {
    sessionId = sid        // MCPClient.swift:228 - no lock held
}
\`\`\`

Two concurrent `callTool` calls on the same connection will race on `sessionId`. Swift 6 strict concurrency will not catch this because of `@unchecked Sendable`, but the race is real.

**Fix**: Protect `sessionId` reads/writes with the same lock (or better: convert `RemoteMCPConnection` to an `actor`; it's all `async` anyway, so there's no stdio-style need for the unchecked escape hatch).

### P1-6: `URLSession.shared` is the wrong choice

`MCPClient.swift:197, 223` use `URLSession.shared`. Two concerns:

1. **Shared cookie jar**: if one remote MCP server sets cookies, they get sent to the next one. Confidential.
2. **Response caching**: sensitive tool outputs may get cached to disk in the system URL cache.

**Fix**: Use a dedicated `URLSession`:
\`\`\`swift
let config = URLSessionConfiguration.ephemeral   // no cookie jar, no disk cache
config.httpAdditionalHeaders = [\"User-Agent\": \"apfel/\\(buildVersion)\"]
config.timeoutIntervalForRequest = TimeInterval(timeoutSeconds)
let session = URLSession(configuration: config)
\`\`\`

### P1-7: Zero Swift unit tests for new code

The PR adds 589 lines of new Swift code and 0 lines of Swift tests. Specifically missing:

- `--mcp-token` flag parsing (happy path, missing value error, empty value, env var precedence)
- `APFEL_MCP_TOKEN` env var handling
- `mcpServerPaths` URL vs path detection
- The scheme validation (after P0-2 fix)
- Bearer-over-http rejection (after P0-2 fix)
- Multiple `--mcp` mixing local + remote

Our project bar (see `CLAUDE.md` release policy and the 322 existing unit tests) is that every new flag has at least one happy-path + one validation test. Please add tests to `Tests/apfelTests/CLIArgumentsTests.swift` following the existing style (tight error assertions: verify `CLIParseError` type AND message content).

### P1-8: `test_round_trip.py` is not wired into integration tests

`mcp/http-test-server/test_round_trip.py` is a manual script that requires three terminal windows to run:
1. Start Python HTTP MCP server
2. Start `apfel --serve --mcp http://...`
3. Run the script

This means CI never exercises the HTTP transport. Zero coverage in automated tests.

**Fix**: Add an integration test to `Tests/integration/` that:
1. Uses a pytest fixture to start the Python HTTP server on a random port
2. Uses the apfel-tests conftest pattern to start `apfel --serve --mcp http://localhost:{port}/mcp`
3. Asserts tool execution round-trips

Follow the pattern in `Tests/integration/mcp_server_test.py` which already does this for the stdio transport.

### P1-9: No visible warning when enabling remote MCP

A user running `apfel --mcp http://evil.example.com/mcp \"summarize my private notes\"` will silently send the prompt + tool arguments to `evil.example.com`. No confirmation, no banner, nothing.

Contrast with the stdio path where the MCP server runs as a local subprocess - it is the user's own code.

**Fix**: When a remote MCP server is configured, print a one-line stderr banner at startup:
\`\`\`
warning: remote MCP server attached (https://example.com/mcp) - tool arguments will be sent to this server
\`\`\`

Goes to stderr so pipelines don't break. Shown always unless `--quiet`.

---

## P2 findings (nice to have)

- **P2-10 scheme defense-in-depth**: `RemoteMCPConnection.init` should verify `url.scheme == \"http\" || url.scheme == \"https\"` independently of the `MCPManager` prefix check, so the class is safe if called from somewhere else in the future.
- **P2-11 per-server tokens**: Currently one `--mcp-token` is sent to ALL remote servers. If server A is compromised, it steals the token intended for server B. Consider `--mcp URL[@TOKEN]` format or mapping tokens to URLs. Acceptable for v1, flag for follow-up.
- **P2-12 User-Agent**: Add `User-Agent: apfel/<version>` so remote MCP operators can identify and rate-limit apfel clients.

---

## What I verified (and what was fine)

- ✅ Build is clean on `feat/remote-mcp` branch (`swift build` green)
- ✅ All 322 existing unit tests pass on the branch (no regressions)
- ✅ `AnyMCPConnection` + `MCPManager` refactor is clean - good abstraction
- ✅ Stdio path is untouched (low regression risk for existing users)
- ✅ SSE parsing is minimal but defensible for the narrow MCP-over-SSE subset
- ✅ `shutdown()` DELETE is per MCP Streamable HTTP spec
- ✅ `URLRequest.setValue` sanitizes header values, so session-ID reflection cannot do CRLF injection
- ✅ Python test server is clean - stdlib only, binds to localhost, no network exfil, no weird dependencies
- ✅ `to_num` helper fix in the calculator servers is a genuine improvement

---

## Suggested path forward

The core design is right. If you are up for it:

1. **Fix P0-1** (APFEL_MCP URL split) - one-line parser fix + test
2. **Fix P0-2** (reject Bearer over http://) - one guard in `post()` + test
3. **Fix P0-3** (response size cap) - switch to dedicated URLSession with config + test
4. **Fix P0-4** (README top-of-file claims) - rewrite the two contradicting lines
5. **Fix P1-5** (sessionId data race) - convert `RemoteMCPConnection` to `actor`, drop `@unchecked Sendable`
6. **Fix P1-6** (dedicated URLSession) - folds into P0-3 fix
7. **Fix P1-7** (unit tests) - add tests to `Tests/apfelTests/CLIArgumentsTests.swift`
8. **Fix P1-8** (integration test) - add `Tests/integration/mcp_remote_test.py` following the stdio pattern
9. **Fix P1-9** (remote warning banner) - 3 lines in `MCPManager.init`
10. **P2 items** can be follow-up PRs if you prefer

If the list is too long, I am happy to take P0-1 + P0-2 + P0-4 as the absolute minimum for merge and land P0-3 / P1-* in a follow-up. But P0-2 (Bearer over http) is a real plaintext-credential leak and must ship as part of this PR.

Let me know which path you prefer and I will re-review quickly. Thank you again for the thoughtful contribution - this is a valuable feature and the architecture you chose makes it straightforward to land once the above is addressed.

---

**Note on the CI status**: GitHub Actions shows \"action_required\" because this is your first PR to the repo - I will approve the CI run shortly so the build-and-test job can verify the branch on `macos-26`.

Bucknalla:
Hi @Arthur-Ficial I've addressed you feedback, thanks!

Arthur-Ficial:
Thanks @Bucknalla - great implementation! I've cherry-picked this to main with additional test coverage and a couple of fixes:

**Changes on top of your PR:**
- Loopback exception in the http:// + bearer token security check (allows `http://127.0.0.1` with a token, which is needed for local dev/testing - only non-loopback http:// is refused)
- URL-aware `parseMCPServerPaths()` helper: handles `APFEL_MCP=https://host:8080/mcp` without mangling the port, supports comma or colon separation, mixes local paths and URLs
- `--mcp-token` added to help text

**Additional tests added:**
- 15 new CLIArguments unit tests (URL parsing, env var precedence, edge cases)
- 17 integration tests covering: bearer token auth (correct token), wrong token at startup (non-zero exit), missing token against auth-required server (non-zero exit), non-loopback http:// + token refused with security message, unreachable server (non-zero exit), SSE streaming with remote MCP tool execution, mixed local stdio + remote HTTP MCP

Released as v0.9.12. `brew update && brew upgrade apfel`

Issue #84, opened 2026-04-13
`make install` doesn't handle non-existent binary dir


aozgaa:
**What happened?**
`make install` tries to elevate at the `install(1)` subcommand If the PREFIX/DESTDIR doesn't exist, then fails anyways:
```
    @if [ -w "$(PREFIX)/bin" ]; then \
      install .build/release/$(BINARY) $(PREFIX)/bin/$(BINARY); \
    else \
      sudo install .build/release/$(BINARY) $(PREFIX)/bin/$(BINARY); \
    fi
```
From `test(1)`:
```
    -w file       True if file exists and is writable.  True indicates only that the write flag is on.  The file is not writable on a read-only file system even if this test indicates true.

```

**What did you expect?**

I'd expect that the install rule either creates the dest dir using inherited permissions, or fails without trying to install/proceed further. In any case, the alternative convention chosen should be documented.

Conventionally, elevating would be taken of by running make itself elevated, something like
```
make build && sudo make install
```
But see https://github.com/Arthur-Ficial/apfel/issues/83 for why this isn't currently practical.

**Steps to reproduce**
1) Case 1 -- PREFIX doesn't exist leads to spurious elevation, then failure anyways
```
% make install PREFIX=${HOME}/.does_not_exist  
0.9.20 → 0.9.21
swift build -c release
[1/1] Planning build
Building for production...
[5/5] Linking apfel
Build complete! (9.95s)
install: /Users/art/.does_not_exist/bin/INS@ogM8iL: No such file or directory
make: *** [install] Error 71
```
2) Case 2 -- bin dir doesn't exist leads to spurious elevation, then failure anyways
 % mkdir -p ~/.does_exist
% make install PREFIX=${HOME}/.does_exist    
0.9.21 → 0.9.22
swift build -c release
[1/1] Planning build
Building for production...
[5/5] Linking apfel
Build complete! (9.43s)
install: /Users/art/.does_exist/bin/INS@ZA8LZD: No such file or directory
make: *** [install] Error 71
```
3) Case 3 -- bin dir exists leads user or admin install based on writeability:
```
% mkdir -p ~/.does_exist/bin                 
% make install PREFIX=${HOME}/.does_exist
0.9.22 → 0.9.23
swift build -c release
Building for production...
[5/5] Linking apfel
Build complete! (4.31s)
✓ installed: apfel v0.9.23

```


**Environment**
- macOS version: (`sw_vers`)
```
% sw_vers
[...]

PR #85, opened 2026-04-13
chore: Install demo scripts to bin

EverWinter23:
## Summary

This change installs the demo scripts to bin dir via make.

The project was quite intriguing. I gave it a go. Unfortunately, the local model is not powerful enough to be useful.
However, it maybe possible that the next set of local models can come in handy in everyday workflows.

```bash
~/workspace/apfel >>> docker image ls                                               
REPOSITORY                       TAG           IMAGE ID       CREATED         SIZE
pgvector/pgvector                pg14          23c2907fbff1   8 months ago    627MB
mongo                            6.0           e6ea23317a13   8 months ago    1GB
redis                            6.2           37440b1f783e   9 months ago    179MB
timescale/timescaledb            latest-pg14   0f5553d37abb   11 months ago   1.68GB
prom/prometheus                  v2.53.4       502ad90314c7   13 months ago   369MB
temporalio/auto-setup            1.24.2        10e8c557ace5   21 months ago   546MB
temporalio/ui                    2.28.0        629b3e444f0b   21 months ago   126MB
rediscommander/redis-commander   latest        19cd0c49f418   4 years ago     133MB

~/workspace/apfel >>> cmd "how do i get all image ids in docker image ls command"                                                                                                                              
$ docker image ls | grep ID

~/workspace/apfel >>> cmd "how do i get the 3rd column of a command output"                                                                                                                                    
$ column -s 3 <output>

~/workspace/apfel >>> docker image ls | column -s 3                                                                                                                                                        
REPOSITORY                       TAG           IMAGE ID       CREATED         SIZE      timescale/timescaledb            latest-pg14   0f5553d37abb   11 months ago   1.68GB
scrollmark-api                   latest        4f41e1fb3f15   6 hours ago     6.6GB     prom/prometheus                  v2.53.4       502ad90314c7   13 months ago   369MB
pgvector/pgvector                pg14          23c2907fbff1   8 months ago    627MB     temporalio/auto-setup            1.24.2        10e8c557ace5   21 months ago   546MB
mongo                            6.0           e6ea23317a13   8 months ago    1GB       temporalio/ui                    2.28.0        629b3e444f0b   21 months ago   126MB
redis                            6.2           37440b1f783e   9 months ago    179MB     rediscommander/redis-commander   latest        19cd0c49f418   4 years ago     133MB

~/workspace/apfel >>> cmd "how do i get the 3rd column of a command output"                                                                                                                                    
$ `echo $1 | head -n 1 | awk '{print $3}'`

~/workspace/apfel >>> docker image ls | echo $1 | head -n 1 | awk '{print $3}'                                                                                                                             
# empty

~/workspace/apfel >>> docker image ls | awk '{print $3}'                                                                                                                                                   
IMAGE
4f41e1fb3f15
23c2907fbff1
e6ea23317a13
37440b1f783e
0f5553d37abb
502ad90314c7
10e8c557ace5
629b3e444f0b
19cd0c49f418                                                                                                                                                                            

~/workspace/apfel >>> explain "docker image ls | awk '{print $3}' "                                                                                                                                        
The command `docker image ls | awk '{print }'` is used to list all Docker images stored in a Docker registry
or local Docker container. The `docker image ls` command retrieves a list of all Docker images present in the
registry or the local Docker environment. The `awk` command is then used to filter out the output, which in
this case is simply printing the image names without any additional formatting.

This command will output a list of all Docker images, each on a new line, but since the `awk` command only
prints the output, it does not modify the output further, and you'll see the image names directly from the
Docker command's output.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,9 @@ PREFIX ?= /usr/local
 BINARY = apfel
 VERSION_FILE = .version
 
-.PHONY: check-toolchain build install uninstall clean bump-patch bump-minor bump-major generate-build-info update-readme version release release-patch release-minor release-major package-release-asset print-release-asset print-release-sha256 update-homebrew-formula preflight benchmark test
+.PHONY: check-toolchain build install uninstall install-demo-scripts uninstall-demo-scripts clean bump-patch bump-minor bump-major generate-build-info update-readme version release release-patch release-minor release-major package-release-asset print-release-asset print-release-sha256 update-homebrew-formula preflight benchmark test
+
+DEMO_SCRIPTS = cmd explain gitsum mac-narrator naming oneliner port wtd
 
 # --- Environment checks ---
 
@@ -202,6 +204,32 @@ uninstall:
 			brew link apfel 2>/dev/null || true; \
 		fi; \
 	fi
+
+install-demo-scripts:
+	@repo_root=$$(pwd); \
+	for script in $(DEMO_SCRIPTS); do \
+		src="$$repo_root/demo/$$script"; \
+		dest="$(PREFIX)/bin/$$script"; \
+		if [ -w "$(PREFIX)/bin" ]; then \
+			ln -sf "$$src" "$$dest"; \
+		else \
+			sudo ln -sf "$$src" "$$dest"; \
+		fi; \
+		echo "✓ linked: $$script → $(PREFIX)/bin/$$script"; \
+	done
+
+uninstall-demo-scripts:
+	@for script in $(DEMO_SCRIPTS); do \
+		dest="$(PREFIX)/bin/$$script"; \
+		if [ -L "$$dest" ] || [ -f "$$dest" ]; then \
+			if [ -w "$(PREFIX)/bin" ]; then \
+				rm -f "$$dest"; \
+			else \
+				sudo rm -f "$$dest"; \
+			fi; \
+			echo "✓ removed: $(PREFIX)/bin/$$script"; \
+		fi; \
+	done
 
 clean:
 	swift package clean


Arthur-Ficial:
Hey @EverWinter23 - thanks for the PR and for actually trying the demos on real workflows. That kind of dogfooding is genuinely useful feedback, even when it surfaces limitations.

After thinking it through, I've decided not to merge this as-is, but I've landed the intent behind it in a different shape. Here's the reasoning:

**Why not auto-install to `$PREFIX/bin`**

1. **Name collisions.** `port` would shadow MacPorts on any machine that has it installed. `cmd`, `explain`, `naming` are generic enough that they could collide with other tools or shell helpers silently.
2. **Symlinks-to-`$(pwd)` are fragile.** If a user `git clone`s, runs `make install-demo-scripts`, then later moves or deletes the repo, they end up with dangling symlinks in their PATH forever.
3. **Silent sudo.** The `if [ -w ... ]; then ln; else sudo ln` branch would fire a sudo prompt mid-`make` with no warning. This was actually the exact pattern we just moved away from in #84 (v1.0.1), where `make install` now fails loudly with a clear error instead of silently escalating.
4. **Demos are demos.** They're intentionally small reference implementations, not shipped commands. The golden goal is three things - `apfel` as a UNIX tool, OpenAI-compatible server, and CLI chat - and the demos are examples of using those, not additional products.

**What I did instead** (commit ccf8e23 on main):

- Added an "Install demos globally (optional)" section to [`demo/README.md`](https://github.com/Arthur-Ficial/apfel/blob/main/demo/README.md#install-demos-globally-optional) documenting the opt-in symlink pattern with an `apfel-` prefix (so `apfel-port 3000`, `apfel-cmd "..."` - no collisions).
- Added a one-line header to each of the 8 demo scripts pointing readers at that section.

The recommended pattern from the README:

```bash
mkdir -p "$HOME/.local/bin"
for d in cmd explain gitsum mac-narrator naming oneliner port wtd; do
  ln -sf "$(pwd)/demo/\$d" "\$HOME/.local/bin/apfel-\$d"
done
```

This gives users who want the demos globally the exact thing your PR was reaching for, without the collision + sudo + dangling-symlink risks of baking it into `make`.

Closing the PR but genuinely appreciated the nudge - the docs improvement wouldn't have happened without it. Feel free to open new PRs anytime.

PR #126, opened 2026-04-24
fix(server): return 200 OK with content_filter for model refusals (#118)

### LINKED ISSUE #118: Populate assistant.message.refusal on 200-OK chat completions (OpenAI wire-format parity)
## Problem

When Apple's on-device model refuses a request, apfel returns it as a `400 Bad Request` with `error.type: "content_policy_violation"`. OpenAI's actual wire format for refusals is **200 OK** with `finish_reason: "content_filter"` and a populated `refusal` field on the assistant message. apfel's encoder hard-codes `refusal` to `null` (`Sources/Core/OpenAIModels.swift:155`), and refusals flow through `ApfelError.refusal -> httpStatusCode 400` (`Sources/Core/ApfelError.swift:112`), so **no amount of classifier work reaches the wire format**.

The recent classifier split (`quality/refusal-preserve-explanation` branch, follow-up to #115/#116) preserves the model's explanation text in `ApfelError.refusal(String)`, but that explanation never reaches the OpenAI response. Clients built against the real OpenAI API see a 400 where they expected a 200 with a structured refusal.

This issue was flagged during the #115 audit and explicitly deferred as a separate scope change.

## Evidence

- `Sources/Core/OpenAIModels.swift:155` - `try c.encodeNil(forKey: .refusal)` unconditionally, even when the caller sets `refusal: "some text"`.
- `Tests/apfelTests/OpenAIWireFormatTests.swift:59-67` - existing test locks in the `null` behavior with comment "Our model never refuses, so the encoder hard-codes refusal=null." That comment is no longer true once `.refusal(String)` exists.
- `Sources/Core/ApfelError.swift:112` - `.refusal` returns HTTP 400, not 200.
- OpenAI spec: `finish_reason` values include `"content_filter"`, paired with a populated `refusal` field on a successful response.

## Six Thinking Hats

**White (facts).** Today: refusals are classified with explanation preserved (in-memory), returned as HTTP 400, OpenAI response body always has `refusal: null`. OpenAI's own servers return 200 + `finish_reason: "content_filter"` + populated `refusal` for the same semantic event. apfel advertises itself as a "drop-in replacement for `openai.OpenAI(base_url=...)`".

**Red (gut feeling).** The current shape feels dishonest to downstream clients. A client that catches `content_policy_violation` today works; a client that inspects `choices[0].message.refusal` like the real API gets nothing useful. Two idioms for one event.

**Black (caution).** This is a **wire-format contract change**. Existing integrations that rely on the 400 path will break silently when upgrading if we flip the semantics. Need a migration plan: flag, deprecation note, or simultaneous dual-emit. The `finish_reason: "content_filter"` taxonomy also needs CLI exit-code review - a refusal on the 200 path should still exit non-zero in the CLI (today: `exitGuardrail = 3` via `Sources/main.swift:32`).

**Yellow (benefits).** Wire-format parity with OpenAI means downstream tooling (LangChain, LlamaIndex, custom clients) Just Works on refusals. The explanation text (already preserved in `ApfelError.refusal(String)`) actually reaches users instead of being lossy-wrapped inside an HTTP error body. Honest about what the on-device model is doing.

**Green (creativity).** Options:
1. **Flag-gated.** `--openai-refusal-200` or `APFEL_OPENAI_REFUSAL_200=1`. Default off for one release, flip to on, remove the flag after two releases.
2. **Surface the async Refusal APIs too.** FoundationModels' `Refusal.explanation` / `explanationStream` are async. To feed the real explanation (not just localizedDescription) we'd need async extraction before `classify()` sees the error. Bigger lift, but cleanly scoped.
3. **CLI unchanged.** Keep exit code 3 (`exitGuardrail`) for refusals even when the HTTP layer returns 200 - CLI semantics and HTTP semantics are separate contracts.

**Blue (process).** TDD: write the wire-format test first (assistant message with refusal populated, `finish_reason: "content_filter"`, status 200), then wire the encoder + handler. Update `OpenAIWireFormatTests.swift:59` assertion. Decide the flag strategy with Franz before touching handlers.

## Technical / Means / Value

**Technical cost.**
- Change `OpenAIMessage.encode(to:)` to emit `refusal` as non-null when set (`Sources/Core/OpenAIModels.swift:141-156`).
- Add a success path in `Sources/Handlers.swift` that detects `ApfelError.refusal`, renders a 200 `ChatCompletion` with the refusal populated on the assistant message and `finish_reason: "content_filter"` on the choice, instead of converting to an HTTP 400 error body.
- Streaming path (`Sources/SSE.swift` / `Server.swift`) needs the same treatment: final chunk with `finish_reason: "content_filter"` and an earlier chunk containing the refusal delta.
- Update `OpenAIWireFormatTests.swift:59-67` plus new integration tests under `Tests/integration/` (requires Apple Intelligence to exercise real refusals).
- CLI unchanged - `exitCode(for:)` at `Sources/main.swift:32` still maps `.refusal` to `exitGuardrail`.

**Means.** Feature flag (env var + CLI flag) for one minor release to let integrations migrate. Document in `docs/openai-compat.md` or similar. Add to STABILITY.md if the flag semantics need a contract.

**Value.** Genuine OpenAI wire-format parity on a previously-missing surface. Preserves the explanation text for downstream clients. Removes the "refusal is always null" quirk that was always a lie about a real refusal semantic, just not exercised until `.refusal(String)` existed.

## Wins and losses

**Wins.**
- Drop-in parity with OpenAI for a real, observable surface.
- Explanation text reaches clients - the thing the classifier work just preserved would actually be used.
- The `refusal` field on `OpenAIMessage` finally means something.

**Losses / costs.**
- Breaking change for any integration that catches HTTP 400 for refusals. Mitigable with a flag and a deprecation window.
- Streaming path complexity grows (one more terminal `finish_reason`).
- Requires Apple-Intelligence-capable Mac for the integration tests - GitHub CI can only cover the unit test deltas.

## Acceptance criteria

- [ ] Non-streaming `/v1/chat/completions`: refusal returns `status 200`, `choices[0].finish_reason: "content_filter"`, `choices[0].message.refusal: "<explanation>"`, `choices[0].message.content: null`.
- [ ] Streaming `/v1/chat/completions`: final chunk carries `finish_reason: "content_filter"` and the refusal text was emitted.
- [ ] `OpenAIMessage.encode(to:)` emits a non-null `refusal` when set; the line 59-67 test in `OpenAIWireFormatTests.swift` is updated.
- [ ] `Tests/integration/test_openai_client.py` (or equivalent) exercises a real FoundationModels refusal and asserts the new shape.
- [ ] CLI behavior preserved: refusal still exits with `exitGuardrail` and prints the `[refusal]` label.
- [ ] Migration story documented (flag, deprecation timeline, STABILITY.md note if needed).

## Context

- Split off from #115 per the closing rationale in the #116 review.
- Depends on (or lands together with) the `quality/refusal-preserve-explanation` branch's `.refusal(String)` classifier work - without that, there's no explanation to surface.
- Does **not** depend on the async `Refusal.explanation` / `explanationStream` FoundationModels APIs for the first cut; those are a nice-to-have second iteration.

cc @franzenzenhofer

---
_Filed by Arthur with Claude Opus 4.7_

Arthur-Ficial:

Fixes #118.

## Root cause

When Apple's on-device model refuses a request, apfel classified it as `ApfelError.refusal(String)` but returned it through the HTTP error path: HTTP 400 with an `OpenAIErrorResponse` payload and `error.type: "content_policy_violation"`. The `OpenAIMessage` encoder also unconditionally hard-coded `refusal: null` (line 155 of `OpenAIModels.swift`), so the refusal explanation text - already preserved in memory by the `.refusal(String)` case - could never reach the wire. OpenAI's actual wire format for refusals is 200 OK with `finish_reason: "content_filter"` and a populated `refusal` field on the assistant message.

## The fix

Six coordinated changes across the data layer, handler layer, and streaming layer:

1. **`OpenAIMessage.encode(to:)`** - The encoder now emits the actual `refusal` value when set and `null` when nil, instead of unconditionally encoding `null`.

2. **`FinishReason.contentFilter`** - New enum case with wire value `"content_filter"`, completing the OpenAI finish_reason taxonomy.

3. **`ApfelError.httpStatusCode`** - `.refusal` now returns 200, not 400. Refusals are successful completions with a content filter, not request errors.

4. **Non-streaming refusal interception** - Both the direct and MCP auto-execute response paths now catch `.refusal` errors and build a proper 200 OK `ChatCompletionResponse` with `finish_reason: "content_filter"`, `message.refusal` populated, and `message.content: null`, instead of routing through the error path.

5. **Streaming refusal interception** - The streaming catch block now emits a `delta.refusal` chunk followed by a `content_filter` finish chunk (plus usage if opted in), instead of an SSE error event.

6. **CLI preserved** - `ApfelExitCodes.code(for: .refusal)` still returns `guardrail` (exit code 3). CLI semantics and HTTP semantics are separate contracts.

No feature flags per Franz's direction - this is a straight bug fix for OpenAI spec compliance.

## Test coverage

- Replaced `OpenAIWireFormatTests` "refusal is always serialized as null, even when set" with two tests: "refusal is serialized as string when set" and "refusal is serialized as null when not set". The old test locked in the broken behavior; the new tests lock in the correct behavior.
- Updated `ApfelErrorTests` "refusal error properties" to expect `httpStatusCode: 200`.
- Added `FinishReason.contentFilter` wire value assertions to `OpenAIWireFormatTests`, `FinishReasonResolverTests`, and `ApfelCorePublicAPIUsageTests`.
- Existing `ExitCodeMapTests` "refusal -> exitGuardrail (3)" is unchanged and continues to lock in CLI behavior.

- Encoder logic follows the existing `content` encoding pattern (conditional nil-check).
- `FinishReason.openAIValue` is an exhaustive switch - adding the new case is compile-safe.
- `Delta` struct gains a `refusal: String?` field with default `nil` in an explicit init, so all existing call sites compile without changes.
- All three handler catch blocks (non-streaming, streaming, MCP auto-execute) intercept `.refusal` before falling through to the error path.
- SSE helpers (`sseRefusalChunk`, `sseFinishChunk`) follow the existing helper pattern.
- No forbidden files touched (no `.version`, `BuildInfo.swift`, CI workflows, release scripts, `Package.swift`).
- Swift 6 concurrency: no data races introduced - all types remain value types.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code or the Swift compiler. If the fix does not compile or tests fail, that is on me to refine - ping me in a review comment.
- Integration tests for real FoundationModels refusals (requires Apple Intelligence).
- Whether the synthesized `Encodable` on `Delta` correctly omits `refusal` when nil in streaming chunks (it should, since synthesized encoding omits nil optionals).

Nothing - straightforward bug fix. The issue was filed by the project's own account and references prior audits (#115, #116). Franz explicitly directed "no flags and fix this as a bug - KISS and TDD!" in the issue comments.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_017YwE1NGCP6eh8XvM9cATsf)_

### DIFF
--- a/Sources/Core/ApfelError.swift
+++ b/Sources/Core/ApfelError.swift
@@ -109,7 +109,7 @@ public enum ApfelError: Error, Equatable, Hashable, Sendable {
     public var httpStatusCode: Int {
         switch self {
         case .guardrailViolation:  return 400
-        case .refusal:             return 400
+        case .refusal:             return 200
         case .contextOverflow:     return 400
         case .rateLimited:         return 429
         case .concurrentRequest:   return 429
--- a/Sources/Core/Chat/FinishReasonResolver.swift
+++ b/Sources/Core/Chat/FinishReasonResolver.swift
@@ -11,6 +11,8 @@ public enum FinishReason: Sendable, Equatable, Hashable, CustomStringConvertible
     case length
     /// The response ended in a tool-call payload instead of plain text.
     case toolCalls
+    /// The response was blocked by the model's content filter (refusal).
+    case contentFilter
 
     /// The OpenAI-compatible wire value for this finish reason.
     public var openAIValue: String {
@@ -18,6 +20,7 @@ public enum FinishReason: Sendable, Equatable, Hashable, CustomStringConvertible
         case .stop: return "stop"
         case .length: return "length"
         case .toolCalls: return "tool_calls"
+        case .contentFilter: return "content_filter"
         }
     }
 
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -113,7 +113,7 @@ public struct OpenAIMessage: Codable, Sendable, Equatable, Hashable {
     public let tool_call_id: String?
     /// Optional sender name.
     public let name: String?
-    /// OpenAI refusal text. Always encoded as `null` for the local model.
+    /// OpenAI refusal text. Populated when the model refuses a request.
     public let refusal: String?
 
     /// Creates an OpenAI-compatible message value.
@@ -151,8 +151,12 @@ public struct OpenAIMessage: Codable, Sendable, Equatable, Hashable {
         try c.encodeIfPresent(tool_calls, forKey: .tool_calls)
         try c.encodeIfPresent(tool_call_id, forKey: .tool_call_id)
         try c.encodeIfPresent(name, forKey: .name)
-        // refusal: always present in responses (null when absent)
-        try c.encodeNil(forKey: .refusal)
+        // refusal: always present in responses (null when absent, string when set)
+        if let refusal = refusal {
+            try c.encode(refusal, forKey: .refusal)
+        } else {
+            try c.encodeNil(forKey: .refusal)
+        }
     }
 
     private enum CodingKeys: String, CodingKey {
--- a/Sources/Handlers.swift
+++ b/Sources/Handlers.swift
@@ -188,6 +188,12 @@ private func mcpAutoExecuteResponse(
         }
     } catch {
         let classified = ApfelError.classify(error)
+        if case .refusal(let explanation) = classified {
+            return refusalResponse(
+                explanation: explanation, id: id, created: created,
+                promptTokens: promptTokens, requestBody: requestBody, events: events
+            )
+        }
         let msg = classified.openAIMessage
         return chatFailure(
             status: .init(code: classified.httpStatusCode),
@@ -322,6 +328,12 @@ private func nonStreamingResponse(
         }
     } catch {
         let classified = ApfelError.classify(error)
+        if case .refusal(let explanation) = classified {
+            return refusalResponse(
+                explanation: explanation, id: id, created: created,
+                promptTokens: promptTokens, requestBody: requestBody, events: events
+            )
+        }
         let msg = classified.openAIMessage
         return chatFailure(
             status: .init(code: classified.httpStatusCode),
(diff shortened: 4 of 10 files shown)


Arthur-Ficial:
Superseded by #127 (merged in v1.3.0). Same fix, same scope - this draft never came out of draft and the new branch landed first. Closing.

The wire-format change (HTTP 200 + finish_reason=content_filter + populated message.refusal) shipped in v1.3.0:
https://github.com/Arthur-Ficial/apfel/releases/tag/v1.3.0

Cheers, Arthur

PR #120, opened 2026-04-24
feat(errors): preserve refusal explanation via ApfelError.refusal(String)

Arthur-Ficial:
## Summary

Preserves the on-device model's actual refusal explanation instead of replacing it with a generic guardrail message.

Today `GenerationError.refusal` collapses into `ApfelError.guardrailViolation`, which has no associated value, so the model's reason is replaced with `"The request was blocked by Apple's safety guardrails. Try rephrasing."` before it ever reaches the CLI or the HTTP error body. This PR threads the explanation through via a new `case refusal(String)`.

### Change

- `ApfelCore`: new `case refusal(String)` on `ApfelError`. Every supporting switch updated (`cliLabel`, `openAIType`, `httpStatusCode`, `openAIMessage`, `debugDescription`, `isRetryable`, `main.swift:exitCode(for:)`).
- `FoundationModelsGenerationErrorCase.refusal` maps to `.refusal(localizedDescription)` instead of `.guardrailViolation`.
- `classifyLocalizedDescription` adds a `refused`/`refusal`/`declined` keyword branch above the guardrail keyword branch.
- Retention semantics: same HTTP 400, same `openAIType: content_policy_violation`, same `isRetryable: false`, same exit code 3 as `guardrailViolation` — script compat preserved. Distinct `[refusal]` CLI label and a message of the form `"The on-device model refused the request: \(explanation)"`.

### Deliberately out of scope (not half-finished)

- **Async `Refusal.explanation` extraction.** `error.localizedDescription` is Apple's locale-rendered surface and carries the explanation in the sync context we already have. No plumbing required.
- **OpenAI wire-format flip from HTTP 400 to 200-with-`message.refusal`.** That's a separate error-vs-success semantic decision; conflating it here would be the half-finished trap. This PR keeps the current error-response shape and only improves the body text.
- **`LanguageModelSession.ToolCallError`.** Confirmed unreachable in apfel's flow: we pass `Transcript.ToolDefinition` (schema-only) to `LanguageModelSession`, never live `Tool` implementations. `ToolCallError` has no throw site.

### API break (by design, semver-minor)

[...]

PR #127, opened 2026-04-24
fix(server): 200 OK + content_filter for on-device refusals (#118)

### LINKED ISSUE #118: Populate assistant.message.refusal on 200-OK chat completions (OpenAI wire-format parity)

Arthur-Ficial:
## Summary
- Fixes #118 as a bug per Franz's direction: 100% OpenAI wire-format compliance, no flag gating, KISS, TDD.
- `ApfelError.refusal` now reports HTTP 200 (output-side filtering is a successful completion per OpenAI spec scenarios 2/5, not an HTTP error).
- `OpenAIMessage.encode(to:)` emits `refusal` as the explanation string when set (was hard-coded to `null`).
- `FinishReason.contentFilter` added with wire value `"content_filter"`.
- `ChatCompletionChunk.Delta` gains a `refusal` field; new SSE helpers emit the refusal delta and the content_filter finish chunk.
- All three handler catch sites (non-streaming, streaming, MCP auto-execute) intercept `.refusal` before the error path and build a 200 response with the refusal populated.

## Scope boundary
- `.guardrailViolation` still returns HTTP 400 - input-side filtering, OpenAI scenario 3. Only `.refusal` flips to 200.
- CLI behaviour unchanged: both `.guardrailViolation` and `.refusal` exit with `ApfelExitCodes.guardrail` (3).

## API break (by design, semver-minor)
Adding `FinishReason.contentFilter` is detected by `swift package diagnose-api-breaking-changes` and the `apfelcore-public-api` CI check correctly fails on this PR. Per [STABILITY.md](STABILITY.md), additive enum cases are a **minor** bump, not a major. Same pattern as #120, which added `ApfelError.refusal(String)` and released as 1.2.0. The red check here is the signal to release as **1.3.0**, not a reason to revert. Existing downstream consumers need to add a `.contentFilter` arm to any exhaustive switch on `FinishReason` - that's the `1.3.0 = new-minor-work` contract.

## Spec reference
Verified against the Azure/OpenAI content-filter scenarios documented at:
- https://learn.microsoft.com/en-us/azure/foundry-classic/foundry-models/concepts/content-filter (scenarios 1-6)
- Vendored OpenAI OpenAPI (`Tests/integration/openai_spec/openapi.yaml`): `finish_reason` enum includes `content_filter`; `choices[].message.refusal` is required-nullable on assistant messages.

## Test plan
- [x] 569 unit tests passing (`swift run apfel-tests`)
- [x] 254 integration tests passing (`make test` - full Apple Intelligence suite, both servers)
- [x] CI `build-and-test` green
- [ ] CI `apfelcore-public-api` **expected red** (API-break on minor bump - see above)

New regression tests:
- `OpenAIWireFormatTests`: `refusal` serialised as explanation string when set; null when nil; control-char escaping; `FinishReason.contentFilter.openAIValue == "content_filter"`
- `ApfelErrorTests` / `ApfelErrorMessageTests`: `refusal.httpStatusCode == 200` locked in
- `FinishReasonResolverTests`: `contentFilter` wire value
- `Tests/integration/openai_client_test.py`: `test_assistant_message_always_exposes_refusal_key` (key always present on assistant messages) and `test_refusal_wire_shape_if_triggered` (regression guard for the 200 + content_filter + populated refusal contract; input-side guardrails treated separately)

### DIFF
--- a/Sources/Core/ApfelError.swift
+++ b/Sources/Core/ApfelError.swift
@@ -106,10 +106,15 @@ public enum ApfelError: Error, Equatable, Hashable, Sendable {
     }
 
     /// HTTP status code for this error type.
+    ///
+    /// `.refusal` returns 200 because an output-side refusal is a successful
+    /// completion per the OpenAI wire format: HTTP 200 with
+    /// `finish_reason: "content_filter"` and the refusal text on the assistant
+    /// message. The CLI exit-code mapping stays separate (`ApfelExitCodes`).
     public var httpStatusCode: Int {
         switch self {
         case .guardrailViolation:  return 400
-        case .refusal:             return 400
+        case .refusal:             return 200
         case .contextOverflow:     return 400
         case .rateLimited:         return 429
         case .concurrentRequest:   return 429
--- a/Sources/Core/Chat/FinishReasonResolver.swift
+++ b/Sources/Core/Chat/FinishReasonResolver.swift
@@ -11,6 +11,9 @@ public enum FinishReason: Sendable, Equatable, Hashable, CustomStringConvertible
     case length
     /// The response ended in a tool-call payload instead of plain text.
     case toolCalls
+    /// The model refused to produce a completion; the refusal text is on the
+    /// assistant message. OpenAI wire value: "content_filter".
+    case contentFilter
 
     /// The OpenAI-compatible wire value for this finish reason.
     public var openAIValue: String {
@@ -18,6 +21,7 @@ public enum FinishReason: Sendable, Equatable, Hashable, CustomStringConvertible
         case .stop: return "stop"
         case .length: return "length"
         case .toolCalls: return "tool_calls"
+        case .contentFilter: return "content_filter"
         }
     }
 
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -113,7 +113,8 @@ public struct OpenAIMessage: Codable, Sendable, Equatable, Hashable {
     public let tool_call_id: String?
     /// Optional sender name.
     public let name: String?
-    /// OpenAI refusal text. Always encoded as `null` for the local model.
+    /// OpenAI refusal text. Populated on assistant messages when the model
+    /// refuses; encoded as null when absent.
     public let refusal: String?
 
     /// Creates an OpenAI-compatible message value.
@@ -151,8 +152,13 @@ public struct OpenAIMessage: Codable, Sendable, Equatable, Hashable {
         try c.encodeIfPresent(tool_calls, forKey: .tool_calls)
         try c.encodeIfPresent(tool_call_id, forKey: .tool_call_id)
         try c.encodeIfPresent(name, forKey: .name)
-        // refusal: always present in responses (null when absent)
-        try c.encodeNil(forKey: .refusal)
+        // refusal: always present in responses (string when the model refused,
+        // null otherwise). OpenAI wire-format parity for content_filter.
+        if let refusal = refusal {
+            try c.encode(refusal, forKey: .refusal)
+        } else {
+            try c.encodeNil(forKey: .refusal)
+        }
     }
 
     private enum CodingKeys: String, CodingKey {
--- a/Sources/Handlers.swift
+++ b/Sources/Handlers.swift
@@ -188,6 +188,21 @@ private func mcpAutoExecuteResponse(
         }
     } catch {
         let classified = ApfelError.classify(error)
+        if case .refusal(let explanation) = classified {
+            if streaming {
+                return await refusalStreamingResponse(
+                    id: id, created: created, promptTokens: promptTokens,
+                    refusal: explanation, includeUsage: includeUsage,
+                    requestBody: requestBody,
+                    events: events + ["refusal: \(classified.cliLabel)"]
+                )
+            }
+            return await refusalNonStreamingResponse(
+                id: id, created: created, promptTokens: promptTokens,
+                refusal: explanation, requestBody: requestBody,
+                events: events + ["refusal: \(classified.cliLabel)"]
+            )
+        }
         let msg = classified.openAIMessage
         return chatFailure(
             status: .init(code: classified.httpStatusCode),
@@ -322,6 +337,13 @@ private func nonStreamingResponse(
         }
     } catch {
         let classified = ApfelError.classify(error)
+        if case .refusal(let explanation) = classified {
+            return await refusalNonStreamingResponse(
+                id: id, created: created, promptTokens: promptTokens,
+                refusal: explanation, requestBody: requestBody,
+                events: events + ["refusal: \(classified.cliLabel)"]
+            )
+        }
         let msg = classified.openAIMessage
         return chatFailure(
             status: .init(code: classified.httpStatusCode),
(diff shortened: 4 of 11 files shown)


Arthur-Ficial (commented):
## Automated First-Pass Review — PR #127

Nice work on this one. The OpenAI wire-format fix is clean, the spec reasoning is solid, and the three-handler interception pattern is consistent. The scope boundary (refusal → 200, guardrail → 400) is exactly right per the Azure/OpenAI content-filter docs.

### What I verified (all clean)

- **Every changed file read in full** (11 files, 338 additions, 25 deletions)
- **`ApfelError.refusal` → HTTP 200**: correct per OpenAI output-side filtering (scenario 2/5). `guardrailViolation` stays 400 (input-side, scenario 3). Scope boundary is precise.
- **`FinishReason.contentFilter`**: new enum case with wire value `"content_filter"`. `openAIValue` switch is exhaustive. `description`/`debugDescription` inherit from `openAIValue` — no missing cases.
- **`OpenAIMessage.encode(to:)`**: refusal is now encoded as the explanation string when set, `null` when nil. Both branches preserve the always-present key. Correct per OpenAI's required-nullable contract on assistant messages.
- **`ChatCompletionChunk.Delta`**: new `refusal` field with default `nil`. Explicit `init` added. All 5 existing call sites pass the original 3 params — default `refusal: nil` applies, no breakage. Two new call sites (`.init(refusal:)` and `.init()`) are correct.
- **Handler interception**: all three catch sites (`nonStreamingResponse`, `streamingResponse` inline, `mcpAutoExecuteResponse`) intercept `.refusal` before `chatFailure`. The refusal path never falls through to the generic error handler, so `chatFailure` is never called with a 200 status for refusals.
- **SSE helpers**: `sseRefusalChunk` streams the refusal delta, `sseContentFilterFinishChunk` emits `finish_reason: "content_filter"` with an empty delta. SSE order is role → refusal delta → content_filter finish → [usage] → [DONE]. Matches OpenAI streaming spec.
- **Streaming inline handler** (Handlers.swift:539–563): correctly yields chunks via continuation rather than delegating to `refusalStreamingResponse`. `completionTokens` is reassigned to count refusal text — used correctly in the completion log at the bottom.
- **`refusalNonStreamingResponse`/`refusalStreamingResponse`**: set `message.content: null` and `message.refusal: <text>`, `finish_reason: "content_filter"`. Token counting on refusal text. Trace events are clean.
- **`SSE.swift` now imports `ApfelCore`**: needed for `FinishReason.contentFilter.openAIValue`. Main target already depends on ApfelCore — no layering violation.
- **No new network, subprocess, auth, or URL-parsing surface** — security audit not triggered.
- **Swift 6 concurrency**: no new `@unchecked Sendable`, no new actor isolation issues. All new types/functions match existing concurrency patterns.

### Test coverage ✓

| Area | Tests | Status |
|------|-------|--------|
| `OpenAIWireFormatTests` | refusal string serialization, null serialization, control-char escaping, `content_filter` wire value | ✅ New |
| `ApfelErrorTests` | `refusal.httpStatusCode == 200` locked in | ✅ Updated |
| `ApfelErrorMessageTests` | `refusal.httpStatusCode == 200` lockdown | ✅ Updated |
| `FinishReasonResolverTests` | `contentFilter.openAIValue == "content_filter"` | ✅ New |
| `openai_client_test.py` | `test_assistant_message_always_exposes_refusal_key` (key always present, null on normal responses) | ✅ New |
| `openai_client_test.py` | `test_refusal_wire_shape_if_triggered` (200 + content_filter + populated refusal text when model refuses) | ✅ New |

### Findings

No P0 or P1 issues.

| Severity | Area | Finding |
|----------|------|---------|
| P2 | `ApfelError.openAIType` | `.refusal` still returns `"content_policy_violation"` — now effectively dead code in the HTTP context since refusals are intercepted before `chatFailure`. Harmless, and the value is still correct if anyone uses the property outside the handler path (e.g., logging). No action needed. |
| P2 | SSE helpers | `sseRefusalChunk` and `sseContentFilterFinishChunk` have no dedicated unit tests. They're thin `ChatCompletionChunk` constructors tested implicitly through the handler tests and the integration suite. Follow-up PR acceptable. |

### Summary

Spec-correct, well-scoped, thoroughly tested. The guardrail/refusal distinction (input-side 400 vs. output-side 200) is the right call per the OpenAI content-filter scenarios. All handler paths are covered, the streaming and non-streaming shapes match the spec, and the test suite locks in the wire-format contract.

CI is in progress (`build-and-test` + `apfelcore-public-api`). Full integration suite (server + model) can only run locally with Apple Intelligence — GitHub CI covers unit tests and model-free integration.

*Automated review — cannot approve or merge. Franz makes the final call.*

---

PR #138, opened 2026-04-28
feat(demo/semantic-cli): add semantic CLI layer with setup, embedding…

aaronroberson:
Adds an optional semantic layer to the demo commands that bridges natural language input with deterministic apfel-* execution.

Instead of choosing between:

* flexible prompts (apfel "explain tcp")
* or fixed commands (apfel-explain tcp)

this layer allows both to work together:

natural language → semantic ranking → canonical apfel-* command

Suggestions are:

* intent-aware (not just prefix-based)
* adaptive (recency + frequency + similarity)
* deterministic at execution (always resolves to apfel-*)

This turns the demo commands into a lightweight, self-learning CLI interface without changing how they execute.

⸻

What’s included

* demo/semantic-cli/setup.sh
    Installs dependencies, symlinks apfel-* commands, and wires zsh hooks
* demo/semantic-cli/embedder.py
    Local embedding + ranking (cosine similarity with recency/frequency weighting)
* demo/semantic-cli/README.md
    Installation + usage
* demo/README.md
    References this as an optional extension

⸻

Design notes

* Fully optional layer (lives under demo/)
* No changes to core apfel code paths
* No dependency on Oh My Zsh or external services
* All execution remains command-based (apfel-*)

⸻

Test plan

* cd demo/semantic-cli && ./setup.sh && exec zsh
* Run apfel explain tcp handshake
* Type apfel expla and verify suggestion resolves to apfel-explain ...
* Confirm no regressions to existing demo commands
* (optional) swift build -c release
* (optional) swift run apfel-tests

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -172,3 +172,57 @@ Then invoke them as `apfel-cmd "find large files"`, `apfel-port 3000`, etc.
 - The symlinks point at your current clone. If you move or delete the `apfel/` directory, the symlinks break - re-run the loop from the new location.
 - Make sure `$HOME/.local/bin` is on your `$PATH` (`echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.zshrc`).
 - To remove later: `for d in cmd explain gitsum mac-narrator naming oneliner port wtd; do rm -f "$HOME/.local/bin/apfel-$d"; done`
+
+⸻
+
+## 🧠 Experimental: Semantic CLI Layer (Optional)
+
+In addition to the existing demo scripts, this repo can be extended with an **alternative semantic, intent-aware layer** that bridges natural language and deterministic commands.
+
+> Note: This is an **optional alternative approach** that complements the existing demo scripts. It does not replace the current demos—both approaches can coexist.
+
+Rather than being limited to either approach, this layer allows both to work together:
+
+> Natural language → semantic understanding → canonical command (`apfel-*`)
+
+### What this enables
+
+- Intent-aware suggestions while typing (`apfel expla` → `apfel-explain ...`)
+- Learning from usage (frequency + recency + similarity)
+- Consistent execution (all suggestions resolve to `apfel-*` commands)
+- Local-only operation (no external API calls required)
+
+### Example
+
+```bash
+apfel explain tcp handshake
+# suggests → apfel-explain tcp handshake
+```
+
+### Status
+
+This is an experimental extension built on top of the demo scripts.
+
+It is not part of core `apfel` yet, but can be layered on via a small install script that adds:
+
+- local embeddings
+- intent routing
+- adaptive suggestions
+
+It is intended as an additive layer, not a replacement for the existing demo workflow.
+
+### How to try it
+
+See `demo/semantic-cli/` for installation and setup (located alongside the demo scripts).
+
+### Why this matters
+
+This turns the demos from standalone scripts into a:
+
+> self-learning command interface
+
+Where:
+- language is used for discovery
+- commands are used for execution
+
+⸻
--- a/demo/semantic-cli/README.md
+++ b/demo/semantic-cli/README.md
@@ -1,0 +1,128 @@
+# Semantic CLI Layer (Experimental)
+
+Adds a semantic, learning-based suggestion layer on top of the apfel demo commands.
+
+Maps natural language input to deterministic `apfel-*` commands that improve with use.
+
+---
+
+## 🚀 Installation
+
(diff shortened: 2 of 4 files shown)


Arthur-Ficial (commented):
Hey @aaronroberson, thanks for sending this over. The intent-routing idea is genuinely nice (natural language → canonical `apfel-*` command) and the Zsh ZLE plumbing in `setup.sh` is well thought through. Before getting into specifics: I'd love to see this exist - just not landed under `demo/` in this repo. Reasoning below, then concrete suggestions.

## Summary findings

| Sev | Area | Finding |
|-----|------|---------|
| P0  | Scope | `demo/` is intentionally tiny one-bash-script-per-thing examples. This is a 594-line system with deps + lifecycle + state. |
| P0  | Positioning | `pip install sentence-transformers` pulls torch (~1-2GB). apfel's pitch is "the free AI already on your Mac" - shipping a demo that asks users to install a parallel ML stack contradicts that. |
| P1  | Side effects | `setup.sh` appends to `~/.zshrc`, creates `~/.zsh/functions/`, writes `~/.apfel_history`. Demos in this repo are read-only invocations - they don't mutate the user's shell config. |
| P1  | Portability | `export PATH="$HOME/Library/Python/3.9/bin:$PATH"` hardcodes Python 3.9. macOS 26 ships 3.13; this line is a footgun. |
| P2  | Style | Emoji headings (🚀 🧪 🧠 ⚙️ 📦) and `⸻` separators don't match the rest of `demo/` which is plain markdown. |

## Why "not in demo/" is the load-bearing call

`demo/` exists to show **apfel** in idiomatic UNIX shape - small scripts that pipe stdin/stdout, each doing one thing, no install steps beyond the `for` loop already in `demo/README.md`. Most of those scripts are 5-30 lines of bash that wrap `apfel "..." `.

What you've built is a **separate product** that wraps the demos: it has its own embedding model, its own state store, its own ZLE hooks, its own history file, its own dependency footprint. The relationship to apfel is "apfel-* commands are the targets that get suggested", not "apfel powers this". That's a great thing for it to be! But it's not a demo of apfel.

If I land this under `demo/`, every future user who runs the existing `for d in cmd explain ...` loop is going to wonder whether they should also install the semantic layer, what it costs, what it changes, whether their `~/.zshrc` is now apfel's responsibility. Today the demo loop is a 30-second thing you can revert with `rm`. Adding this changes that contract.

## Concrete suggestions

1. **Split it into its own repo** - `apfel-semantic-cli` or similar. You own it, you ship updates on your cadence, no dependency on apfel's release rhythm. I'm happy to add it to `README.md`'s [Community Projects](https://github.com/Arthur-Ficial/apfel#community-projects) section when it's live, the same way [apfel-for-raycast](https://github.com/raycast/extensions/pull/26954) gets called out. That's the apfel-ecosystem story we want.

2. **Drop the heavy ML dep if you can** - apfel itself runs Apple Intelligence on-device. If you point your embedder at apfel's own `/v1/chat/completions` (or a future `/v1/embeddings` if Apple ever ships embeddings via FoundationModels), you keep the "no extra ML stack" promise. Today apfel returns 501 on `/v1/embeddings`, so you'd need a different on-device approach (e.g. macOS NLEmbedding via Swift, or a small `mlx-embeddings` model). Worth thinking about - "100% on-device including the routing layer" would be a strong story for your project.

3. **Make the install opt-in and reversible** - even outside this repo, `setup.sh` should ship a `setup.sh --uninstall` and should never blindly append to `~/.zshrc`. Common pattern: write to `~/.zsh/apfel-semantic.zsh`, instruct the user to add a single `source` line themselves. That keeps user shell config under user control.

4. **Don't hardcode Python paths** - use `python3 -m pip --user` and let the user's existing `PATH` resolution handle binaries. The `Library/Python/3.9/bin` line will silently break on every macOS upgrade.

## What I'd happily merge

A small (≤30 line) `demo/intent-router.zsh` that reads `~/.zsh_history`, looks at the prefix the user typed, and prints the closest `apfel-*` command via levenshtein or simple word overlap - no embeddings, no Python, no `~/.zshrc` mutation. Pure shell, fits next to the other demo scripts. Happy to take that PR.

## Closing this PR (with thanks)

Going to close as **not merging here**, but please open the standalone repo - I'll genuinely link it once it's live. The naming-router intuition is solid; the wrapper-around-demos shape is the part that needs to live elsewhere.

If I've misread the intent and you do want to slim it down to fit `demo/`, say the word and I'll reopen.

Cheers,
Arthur

PR #151, opened 2026-05-16
feat: Detect a MacPorts installation on update

halostatue:
I'm adding a MacPorts port for apfel and want to indicate that the source is MacPorts. Unlike a Homebrew upgrade, this should _not_ be managed by the binary itself and its instructions should be accurate for MacPorts.

Until this is merged, we'll be applying a simple patchfile which replaces the whole function with the print statements.

Developed with the assistance of Kiro using Claude Opus 4.7.

### DIFF
--- a/Sources/CLI.swift
+++ b/Sources/CLI.swift
@@ -317,9 +317,23 @@ func performUpdate() {
     if isBrew {
         print("\(appName) v\(current) (installed via Homebrew)")
     } else {
-        print("\(appName) v\(current) (installed from source)")
-        print("To update: git pull && make install")
-        print("Or visit: https://github.com/Arthur-Ficial/apfel/releases")
+        // Detect MacPorts by walking up from the binary to ${prefix}/var/macports.
+        // Default prefix is /opt/local; this works for custom prefixes too.
+        let macportsURL = URL(fileURLWithPath: resolved)
+            .deletingLastPathComponent()  // ${prefix}/bin
+            .deletingLastPathComponent()  // ${prefix}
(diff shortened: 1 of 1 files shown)


Arthur-Ficial:
Thanks @halostatue - this landed on `main` as [401f244](https://github.com/Arthur-Ficial/apfel/commit/401f244) with you preserved as Co-Author.

I made two follow-up tweaks while merging:

1. **Pulled the detection logic into a pure helper** at `Sources/CLI/InstallMethod.swift` (`enum InstallMethod` + `func detectInstallMethod(binaryPath:)`). This lives in `ApfelCLI` so it has no FoundationModels dependency and is straightforward to unit-test.
2. **Added 8 unit tests** in `Tests/apfelTests/InstallMethodTests.swift` covering Homebrew (Cellar, opt symlink, Intel `/usr/local/homebrew`), MacPorts default `/opt/local` prefix, MacPorts custom prefixes, the file-not-directory edge case, and source-built paths. Test count: 599 -> 607, all green.

The behaviour is identical to your patch: `${prefix}/var/macports` exists as a directory -> "installed via MacPorts" + `sudo port sync && sudo port update apfel`.

Closing this PR; the change will ship in the next release (cutting v1.3.6 shortly). Appreciate the contribution and the clear write-up - the MacPorts portfile is exactly the kind of distribution channel we want covered properly.

Cheers, Arthur

cc @franzenzenhofer

---
_Reviewed and merged by [Claude Code](https://claude.ai/code)_

halostatue:
Verified with the update to 1.3.8.

PR #157, opened 2026-05-21
Add --context-status Strategy to print context fill after each turn

BogdanGeorgian91:
## Summary
`--context-status` prints the current chat context fill after each turn, for example `[context 2381/3584 tokens, 66%, 1203 remaining]`. It uses the same token count and input budget that apfel already checks before rotating context.

## Test plan
- [ ] `swift build -c release`
- [ ] `swift run apfel-tests`
- [ ] `make install` and manual smoke test

PR #159, opened 2026-05-21
feat(chat): add built-in developer tools

BogdanGeorgian91:
## Summary

This PR adds built-in developer tooling around apfel's existing on-device model workflow. It makes the helper scripts usable from the main `apfel` binary, exposes them inside interactive `--chat` sessions through slash commands, and documents the execution model so users can understand when a request goes to the model versus when apfel runs a local tool.

The main user-facing change is that developer utilities such as `cmd`, `explain`, `wtd`, `port`, `process`, `daemon`, `docs-apple`, and `mdn` can now be invoked directly through apfel instead of requiring manual symlinks or separate demo-script setup.

### Chat slash commands

- Added `ChatCommand`, a pure parser for slash commands in `--chat` mode.
- Added session controls: `/clear`, `/new`, and `/context`.
- Added slash-command dispatch for developer tools: `/cmd`, `/oneliner`, `/naming`, `/explain`, `/wtd`, `/port`, `/process`, `/daemon`, `/docs-apple`, and `/mdn`.
- Kept bare words in chat routed to the model, so ordinary prompts such as `explain recursion` or `port is a networking concept` do not accidentally trigger tools.
- Added argument handling that preserves shell snippets, strips double quotes used for grouping, preserves single quotes, and keeps leading flags separate from the main query.

### Built-in developer-tool execution

- Added a fast path in `main.swift` for direct terminal usage such as `apfel port 3000`, `apfel explain "awk -F: '{print $1}' /etc/passwd"`, and `apfel docs-apple SwiftUI Button`.
- Added `EmbeddedScripts`, which packages the developer scripts into the binary and extracts them to `~/.apfel/bin` when used.
- Added script extraction for helper files used by `docs-apple` and `mdn`.
- Added safeguards in chat mode that block interactive flags such as `-x`, `--execute`, `-c`, and `--copy`, because those require direct terminal control and can leave the chat line editor in a bad state.
- Re-injects slash-command output into the chat transcript, so follow-up questions can refer to tool results.

### New and expanded developer tools

- Added `demo/process` to explain a process by PID using local process metadata and on-device model reasoning.
- Added `demo/daemon` to explain macOS daemons and system services.
- Added `demo/docs-apple`, an Apple documentation helper that searches/fetches Apple developer documentation before falling back to model-only answering.
- Added Apple docs parser and context-trimming helpers under `demo/helpers/`.

### Command behavior details

- `/docs-apple <query>` searches Apple developer documentation from chat. It builds a focused docs query, prefers `sosumi search` and `sosumi fetch` when the `sosumi` CLI ([https://sosumi.ai/](https://sosumi.ai/)) is installed, and falls back to native `curl` plus the bundled DocC parser when `sosumi` is unavailable or `--no-sosumi` is passed.
- `/docs-apple` trims fetched DocC content before model invocation. It drops navigation-heavy sections such as relationships, conforming types, inherited members, and see-also blocks, while keeping high-value sections such as overview, declaration, discussion, parameters, creating, and using sections.
- `/mdn <query>` searches MDN Web Docs ([https://developer.mozilla.org/](https://developer.mozilla.org/)) directly through MDN's public documentation API. It does not require or start an MCP server. It searches MDN, fetches the top result's JSON documentation payload, extracts useful prose, and trims low-signal material such as specification tables, browser-compatibility data, and see-also link lists.
- `/clear` clears the terminal display but keeps the current conversation context.
- `/new` clears the terminal display and starts a fresh model session while preserving the configured system prompt and tool instructions.
- `/context` prints the current input-context usage on demand, using the same token budget apfel uses internally after reserving output tokens.
- `@keyword` forces exact docs search terms for `/docs-apple` and `/mdn`; budget flags such as `--1000`, `--2000`, and `--3000` tune how much fetched documentation is returned.

### Why script dispatch instead of MCP

- The chat commands use direct script dispatch so apfel can capture each command's output and inject it back into the same chat transcript. That preserves context for follow-up questions such as "explain the result" or "use that API in an example".
- MCP tools are useful for externally supplied tools, but using MCP for these built-in helpers would add another tool-registration and tool-call layer for functionality apfel already owns. The slash-command path keeps the behavior deterministic: `/docs-apple` and `/mdn` run the local helper, return text, and append that text to the current session.
- This also avoids starting and supervising a separate local MCP server for built-in commands. The scripts remain portable because `EmbeddedScripts` extracts them from the installed binary when needed.
- The result is simpler for users and better for chat continuity: local documentation lookup happens immediately, and the model receives the grounded result as part of the ongoing conversation rather than as a separate MCP interaction.

### Documentation updates

- Documented chat slash commands in `README.md` and `docs/cli-reference.md`.
- Updated `docs/context-strategies.md` to describe the new context-status behavior and `/context` command.
- Expanded `demo/README.md` and `docs/demos.md` with direct invocation, chat-mode usage, execution lifecycle, quoting rules, and docs-search behavior.

### Tests

- Added `ChatCommandsTests` for the pure slash-command parser.
- Added regression coverage for case-insensitive session commands, false-positive prevention, quote handling, shell-snippet preservation, leading flag separation, `docs_apple` normalization, `/mdn`, and `@keyword` parsing.
- Added `DocsAppleScriptTests` to lock down the docs-first behavior in `demo/docs-apple`.
- Wired the new test suites into the custom test runner.

## Why

The helper tools are useful, but their previous demo-script workflow made them harder to discover and harder to use from apfel's primary surfaces. This change keeps apfel's core pitch intact: the binary remains a UNIX tool and an OpenAI-compatible server, while the chat UI gains practical local shortcuts for quick development tasks.

The slash-prefix design keeps chat predictable. A normal sentence goes to the model. A slash command runs a local tool. That separation avoids accidental dispatch and makes the interaction model clear.

## Verification

- Ran `swift run apfel-tests` locally.
- Result: all 662 tests passed.


Arthur-Ficial (commented):
Hi @BogdanGeorgian91, first off — thank you for the depth of work here. The slash-command parser is genuinely careful: quote handling, leading-flag separation, case-insensitive matching, the `@keyword` and budget-flag syntax. The 600+ tests passing locally, the dedicated `ChatCommandsTests` and `DocsAppleScriptTests` suites, the re-injection of tool output into the chat transcript so the model can reference it in follow-ups — that's all thoughtful, well-tested work.

I want to be honest with you up front though, because I'd rather not waste your time: there are two things I want to flag — one is a hard correctness issue, and the other is a direction-fit question that's ultimately Franz's call, not mine. I'll lay both out, and then suggest a path forward that we'd genuinely love to support.

## The one hard correctness issue

The new fast path in [`Sources/main.swift:57-69`](https://github.com/Arthur-Ficial/apfel/pull/159/files) currently hijacks ordinary prompts:

```swift
if let firstWord = parts.first,
   ["cmd", "oneliner", "naming", "explain", "wtd",
    "port", "process", "daemon", "docs-apple", "docs_apple"].contains(firstWord) {
    ...
    let (_, status) = runDeveloperTool(tool: firstWord, arguments: toolArgs, captureOutput: false)
    exit(status)
}
```

Because the check is purely on the first whitespace-split word of `argv[1]`, a user running any of these legitimate prompts gets routed into a dev-tool subprocess instead of the model:

```bash
apfel "explain quantum entanglement"        # → 'explain' fast path
apfel "process the data and summarise"      # → 'process' fast path (expects a PID)
apfel "port the python script to swift"     # → 'port' fast path (expects a number)
apfel "naming considerations for APIs"      # → 'naming' fast path
apfel "daemon worship in early religion"    # → 'daemon' fast path
apfel "cmd-shift-z is redo on mac"          # → 'cmd' fast path
```

These are all common English starters for prompts — apfel today happily answers all of them. After this PR they'd silently break or produce confusing output. Worth addressing regardless of where the rest of the PR lands — a required leading marker (e.g. an explicit subcommand prefix like `apfel run cmd ...`) would resolve it.

## The direction-fit question (Franz's call)

apfel's [`CLAUDE.md`](https://github.com/Arthur-Ficial/apfel/blob/main/CLAUDE.md) is fairly explicit about the project's two core products being the UNIX tool and the OpenAI-compatible server, with chat described as "a byproduct for quick testing, not a main product" and the README structure rule asking chat to stay a short subsection. This PR does grow chat into a richer surface (10 slash commands, embedded tool runtime, expanded README table covering the slash language) and adds outbound HTTP from the main binary (`docs-apple` calls `developer.apple.com`, `mdn` calls MDN's API).

That isn't a verdict from me — it's a tradeoff Franz needs to weigh against the project direction. I want to be transparent that this part of the review reflects a project-direction question, not a code-quality one, and the call belongs to Franz.

## A path forward we'd love to support

If you're up for it, the work in this PR would be a great fit for **its own sister project** — something like `apfel-tui` or `apfel-dev-tools` (pick whatever name feels right to you). The model would be:

- **`apfel`** stays the small, sharp engine — UNIX tool + OpenAI-compatible server.
- **`apfel-tui`** (or whatever you call it) sits on top of `apfel`, owns the rich TUI / slash-command / dev-tooling experience, and uses `apfel --serve` or `apfel`-the-CLI underneath as the on-device-AI backend. You get full creative freedom — your roadmap, your release cadence, your design choices, no upstream golden-goal constraints.

We'd be genuinely happy to:

- **Link to it from the apfel README** in the "Built on apfel" section alongside the other ecosystem projects (`apfel-chat`, `apfel-quick`, `apfel-clip`, `apfelpad`, etc.), so users can find it.
- **Contribute back** to it where it makes sense — bug fixes, ideas, integration testing against new apfel releases.
- Help with the practical bootstrap (Homebrew tap entry, README scaffolding, CI setup) if you'd like a hand getting it off the ground.

Your hand on the wheel, our public endorsement and contribution from the side. The architecture in this PR (clean parser separation, capture-and-reinject transcript model, embedded scripts approach) is genuinely well-built — it deserves a home where it can grow without competing for room with apfel's main pitch.

## Smaller correctness notes (independent of direction)

These are worth fixing wherever this work eventually lands.

- `Sources/CLI.swift` `printContextStatus` regresses the format that just landed in [#157](https://github.com/Arthur-Ficial/apfel/pull/157): the percent, remaining-tokens, styled brackets, and `.dim` styling all disappear. Worth restoring.
- Three independent lists of tool names exist (`main.swift` fast path, `runDeveloperTool` whitelist, `ChatCommand.toolNames`) and have already drifted slightly — `mdn` is in two of three. A single `static let toolNames` would prevent the drift.
- The PR was opened from your fork's `main` branch rather than a feature branch — cosmetic only, but a feature branch keeps your `main` clean for follow-up PRs.

## tl;dr

- **The work itself is well-built** — parser, tests, chat re-injection design are good.
- **One hard bug** (prompt hijacking on common English first-words) should be fixed independent of anything else.
- **Direction-fit is Franz's call**, not mine — flagging it transparently rather than prescribing.
- **The path we'd love to see**: a sister project (`apfel-tui` / `apfel-dev-tools` / your name) sitting on top of apfel. We'd link to it from the README and contribute back; you keep full ownership and creative control.

Thanks again for the work — let me know which direction sounds good and we'll go from there.

Cheers,
Arthur

BogdanGeorgian91:
Copy, I had a feeling you might say that :) totally understand. 
thank you for the review, feedback and availability to help in the future. 
I'm considering what you've said as an alternative to add the integrations I had in mind and some others. keep in touch i guess

Arthur-Ficial:
Thanks for the gracious reply, @BogdanGeorgian91 — and for the work itself, which really is well-built. Going to close this PR on the path we discussed, but I want to leave the door wide open for what comes next.

## Where this lands

- **PR closed** against `apfel` per the direction-fit + the hard prompt-hijacking bug in `Sources/main.swift:57-69`.
- **The work is not wasted.** The clean slash-command parser, the capture-and-reinject transcript model, the embedded-scripts approach, the 600+ tests — all of that is solid foundation for a sister project that owns the rich TUI / dev-tooling experience without competing with apfel's UNIX-tool + OpenAI-server pitch.

## If you want to spin up the sister project

Pick the name (`apfel-tui`, `apfel-dev-tools`, `apfel-slash` — your call), and ping me when the repo exists. I'm happy to:

- Add it to the "Built on apfel" section of the `apfel` README alongside the other ecosystem projects.
- Help with the Homebrew tap entry (`Arthur-Ficial/homebrew-tap`) so users can `brew install Arthur-Ficial/tap/<your-name>`.
- Contribute back on integration testing whenever a new `apfel` release ships, so your tool keeps working as the engine evolves.
- Review early architecture sketches if you want a second pair of eyes.

You'd own it end-to-end — your roadmap, your release cadence, no upstream golden-goal constraints. The ecosystem (apfel-chat, apfel-quick, apfel-clip, apfelpad, apfel-server-kit) is built exactly this way: small sharp engine in the middle, opinionated tools on top.

Thanks again for the depth and the patience. Looking forward to seeing what you build.

Cheers,
Arthur

Issue #119, opened 2026-04-24
Park: add ToolCallError classifier branch only when apfel registers live Tool implementations


Arthur-Ficial:
## Problem (parked)

`FoundationModels.LanguageModelSession.ToolCallError` is a real SDK error type that wraps a tool plus an underlying error. `ApfelError.classify(_:)` has no explicit typed path for it. That sounds like a gap, but it is not: **the error type is unreachable in apfel's current architecture**, so adding a classification path would be error handling for a scenario that cannot happen.

This issue parks the question. File as `wontfix-until-unblocked` / `parked` - reopen **only** when apfel actually starts registering live `Tool` implementations with FoundationModels.

## Why it is unreachable today

FoundationModels throws `ToolCallError` when a `Tool` **implementation** it owns throws during execution. apfel never hands implementations to FoundationModels:

- `Sources/SchemaConverter.swift` produces `Transcript.ToolDefinition` values - **schema only**.
- `Sources/ContextManager.swift` and `Sources/Session.swift` (`makeSession`, `makeTranscriptSession`) assemble transcripts with `toolDefinitions: []` or definition-only tool blocks. Never `tools: [SomeTool()]`.
- The tool loop in `Sources/Session.swift` (`detectAndExecuteMCPTools`, `executeMCPToolCallsForCLI`, `executeMCPToolCallsForServer`) is apfel-side: the model emits a tool-call request in its output, apfel parses it (`ToolCallHandler.detectToolCall`), runs the tool via `MCPClient`, feeds the result back as a tool-role message, and continues generation. FoundationModels is never the thing that calls the tool.
- Errors from that apfel-side loop surface as `MCPError`, which `ApfelError.classify(_:)` already handles explicitly at `Sources/Core/ApfelError.swift:20-22` and maps to `.toolExecution(...)`.

Net: no code path in apfel today can produce a `ToolCallError`. A classifier branch for it would be dead code.

## Six Thinking Hats

**White (facts).** apfel's tool architecture is out-of-band: schema in, structured text out, apfel invokes the tool, feeds the result back. FoundationModels' native `Tool` protocol and in-band invocation are not used. `ToolCallError` is thrown by the native path only.

**Red (gut feeling).** Tempting to "just add the branch for safety" - CLAUDE.md specifically warns against this: "Don't add error handling for scenarios that can't happen." Adding it now would be comfort code that rots.

[...]

PR #200, opened 2026-06-09
Handle function-name string tool calls

dijs:
## Summary

  Accept tool call payloads where `function` is emitted as a string and
  `arguments` is provided as a sibling field. This lets MCP auto-execution
  handle an additional FoundationModels tool-call shape instead of leaking
  raw tool_calls JSON back to clients.

## Test plan
- [x] `swift build -c release`
- [x] `swift run apfel-tests`
- [x] `make install` and manual smoke test

### DIFF
--- a/Sources/Core/ToolCallHandler.swift
+++ b/Sources/Core/ToolCallHandler.swift
@@ -231,13 +231,25 @@ public enum ToolCallHandler {
 
         var result: [ParsedToolCall] = []
         for call in rawCalls {
-            guard let id = call["id"] as? String,
-                  let fn = call["function"] as? [String: Any],
-                  let name = fn["name"] as? String else { continue }
+            guard let id = call["id"] as? String else { continue }
+
+            let name: String
+            let rawArguments: Any?
+            if let fn = call["function"] as? [String: Any],
+               let fnName = fn["name"] as? String {
+                name = fnName
+                rawArguments = fn["arguments"]
+            } else if let fnName = call["function"] as? String {
+                name = fnName
+                rawArguments = call["arguments"]
+            } else {
+                continue
+            }
+
             let args: String
-            if let s = fn["arguments"] as? String {
+            if let s = rawArguments as? String {
                 args = ensureJSONArguments(s)
-            } else if let obj = fn["arguments"],
+            } else if let obj = rawArguments,
                       let data = try? JSONSerialization.data(withJSONObject: obj),
                       let s = String(data: data, encoding: .utf8) {
                 args = s
--- a/Tests/apfelTests/ToolCallHandlerTests.swift
+++ b/Tests/apfelTests/ToolCallHandlerTests.swift
@@ -48,6 +48,13 @@ func runToolCallHandlerTests() {
         try assertNotNil(result)
         try assertEqual(result!.first?.argumentsString, "{\"key\":\"val\"}")
     }
+    test("parses function name string shape with sibling arguments") {
+        let response = #"{"tool_calls": [{"id": "call_unique1", "type": "function", "function": "get_current_conditions", "arguments": "{\"latitude\": 37.7749, \"longitude\": -122.4194}"}]}"#
+        let result = ToolCallHandler.detectToolCall(in: response)
+        try assertNotNil(result)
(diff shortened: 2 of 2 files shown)


Arthur-Ficial (commented):
## Review - PR #200: Handle function-name string tool calls

First: thank you - this is exactly the kind of contribution apfel loves. A real-world FoundationModels output quirk, a surgical fix in the right layer (`ApfelCore`'s tolerant out-of-band parser), a regression test in the right file, and a filled-in test plan. The diff consolidates the arguments handling for both shapes instead of duplicating it, which is nicer than the minimum fix.

### Findings

| # | Priority | Area | Summary |
|---|----------|------|---------|
| 1 | P2 | Tests | Missing edge tests for the new shape (absent sibling `arguments`; non-dict/non-string `function`) |

No P0/P1. Details:

**P2 - edge tests for the new shape.** The repo standard is that every parse branch has its edges pinned. Two cases worth a test each in `Tests/apfelTests/ToolCallHandlerTests.swift`:

1. `function` string with **no** sibling `arguments` -> should parse with `argumentsString == "{}"`:
```swift
test("function name string shape without arguments defaults to empty object") {
    let response = #"{"tool_calls": [{"id": "c1", "type": "function", "function": "get_time"}]}"#
    let result = ToolCallHandler.detectToolCall(in: response)
    try assertNotNil(result)
    try assertEqual(result!.first?.argumentsString, "{}")
}
```
2. `function` that is neither a dict-with-name nor a string (e.g. a number) -> call skipped (documents the `continue` path).

Happy to take these in a follow-up if you prefer - they should not block this fix.

### Golden-goal / architecture check (all clean)

- **Out-of-band tool-calling boundary respected**: no `FoundationModels.Tool` conformance, no native in-band registration - parsing only. #119 stays parked; no `ApfelError` classifier branch needed.
- **Right layer**: pure `ApfelCore`, no new dependencies, unit-testable - and unit-tested.
- **Spec compliance improves**: the variant shape is accepted on *input* (model output), while everything downstream still emits the canonical OpenAI nested `function` object - clients never see the variant.
- **Closes a real leak**: previously this shape made `detectToolCall` return nil, so raw `tool_calls` JSON passed through to users in MCP auto-execution (CLI and server paths).
- **No new attack surface**: arguments already flowed from model output through `ensureJSONArguments`/JSON re-serialization; the new branch reuses the exact same sanitization. No force-unwraps, no `@unchecked Sendable`, no behavior change for well-formed outputs.

### What I verified locally

- Read the full `ToolCallHandler.swift` post-patch and traced every branch combination (dict-without-name, string `function` + string/object/missing/absent `arguments`, non-string `id`) - all degrade exactly like the old code (skip or `{}`).
- `swift build` clean on the PR head.
- `swift run apfel-tests`: **all 653 tests pass** (including your new test).
- Full integration suite against the PR's release binary with live Apple Intelligence + MCP calculator server: **271/272 pass**. The single failure (`test_man_page.py::test_version_matches_version_file`) is environmental - your branch is based on pre-1.5.3 (`.version` 1.5.2) while this machine has 1.5.4 installed; the same test passes on current main. Not caused by this PR.
- CI runs approved (first-time contributor) and in progress.

### Suggested path forward

Mergeable as-is once CI is green; the P2 edge tests can land here or as a follow-up. The final merge call is Franz's (repo policy).

*Review prepared by Arthur Ficial with Claude Code (session model: claude-fable-5).*

Arthur-Ficial:
Merged and **released** - thank you, @dijs! 🎉

Your fix shipped in [v1.5.5](https://github.com/Arthur-Ficial/apfel/releases/tag/v1.5.5) about ten minutes after merge:

- Squash-merged with your authorship preserved (`Co-Authored-By`)
- Added the two follow-up edge tests from the review (absent sibling `arguments` -> `{}`, object sibling arguments serialized, non-dict/non-string `function` skipped)
- Full qualification before release: 661 unit + 274 integration tests, 0 skipped, against live Apple Intelligence

**Get it:** `brew upgrade apfel` (tap is live now; homebrew-core autobump lands within ~24h, nixpkgs follows).

This was a model first contribution: real on-device quirk, minimal fix in the right layer, regression test included. The door is wide open for more. 🍎

PR #203, opened 2026-06-10
fix(homebrew): drop redundant `depends_on :macos` from tap formula

johnpaulashenfelter:
## Summary

`brew install Arthur-Ficial/tap/apfel` (and any `brew` operation that loads the formula) emits:

> Calling `depends_on :macos` with `depends_on macos:` is deprecated!

The tap formula declares both `depends_on :macos` and `depends_on macos: :tahoe`. The version form already implies macOS, so the bare requirement is redundant. `brew style` flags it as `Homebrew/OSDependsOn: Remove redundant depends_on :macos` (autocorrectable).

The tap's `Formula/apfel.rb` is generated by `scripts/write-homebrew-formula.sh` on every `make release`, so the fix belongs at the generator source — editing the tap directly would be overwritten on the next release. This drops the redundant line; `depends_on macos: :tahoe` continues to enforce the macOS 26 Tahoe minimum.

Scope: only the `Arthur-Ficial/homebrew-tap` formula is generated from this script. homebrew-core maintains its own formula (autobump only touches url/sha256) and is unaffected.

## Test plan

- `bash scripts/write-homebrew-formula.sh --version 9.9.9 --sha256 <64-zeros> --output /tmp/apfel.rb`
- `brew style /tmp/apfel.rb` — the `Homebrew/OSDependsOn` offense is gone (the remaining generic cops only fire when linting a standalone file outside a tap).

### DIFF
--- a/scripts/write-homebrew-formula.sh
+++ b/scripts/write-homebrew-formula.sh
@@ -41,7 +41,6 @@ class Apfel < Formula
   license "MIT"
 
   depends_on arch: :arm64
-  depends_on :macos
   depends_on macos: :tahoe
 
   def install


tomjn:
@Arthur-Ficial anything we can do to help get this morning?

tomjn (approved):

Arthur-Ficial (approved):
Thanks for this, @johnpaulashenfelter - clean, correctly-scoped fix and a genuinely good first contribution to apfel. You found the right root cause (the generator, not the generated tap file, which would just be overwritten on the next `make release`), you scoped it precisely, and your test-plan reproducer is spot on. Appreciated.

## Findings

| Severity | Area | Summary |
|----------|------|---------|
| - | correctness | None. Change is correct and safe. |

## What I verified

- **The deprecation is real.** `depends_on :macos` alongside `depends_on macos: :tahoe` triggers Homebrew's `Calling depends_on :macos with depends_on macos: is deprecated!` warning and the `Homebrew/OSDependsOn` cop (autocorrectable). The version form already implies macOS.
- **Linux is still blocked after the change.** This was my one concern - apfel must never be installable on Linux. `depends_on macos: :tahoe` requires macOS 26 Tahoe; no Linux system can satisfy a macOS *version* requirement, so dropping the bare `depends_on :macos` does not open a Linux install path. The arm64 + Tahoe constraints remain intact.
- **Scope is correct.** Only `scripts/write-homebrew-formula.sh` (the `Arthur-Ficial/homebrew-tap` generator) is touched. homebrew-core maintains its own formula and its autobump only rewrites url/sha256, so it's unaffected - exactly as your PR body states.
- **Generated formula still builds the demo companion commands and the service block** - I regenerated with a dummy version/sha and diffed; only the redundant line is gone.

## Path forward

Approving. This takes effect only on the next `make release` (when the generator runs), so merging publishes nothing on its own. Leaving the actual merge + release timing to @franzenzenhofer per our process.

One optional follow-up (not blocking, not yours to fix here): homebrew-core's `apfel.rb` may carry the same redundant line - if so that's a separate one-line PR to Homebrew/homebrew-core, since we don't generate that formula.

Arthur-Ficial:
@johnpaulashenfelter - first, thank you: you flagged a real warning and you found the right layer to fix it (the generator, not the generated file). I approved this initially, but on deeper verification I have to walk that back, and I owe you the reasoning.

## Why dropping the line is unsafe *for this formula*

The `Homebrew/OSDependsOn` cop is right in general, but it's a **false positive for our tap**, because the tap is a prebuilt-binary formula, not a build-from-source one.

Compare the two formulae:

- **homebrew-core `apfel.rb`** has `depends_on xcode: ["26.4", :build]` and builds with `swift build`. Xcode can't exist on Linux, so Linux is excluded structurally - core genuinely doesn't need `depends_on :macos`.
- **Our tap formula** has no xcode build-dep; it does `bin.install "apfel"` from a `...-arm64-macos.tar.gz`. Its only OS guards are `arch: :arm64` and `macos: :tahoe`.

I tested what `depends_on macos: :tahoe` alone actually does, in a real tap context:

```
# version form only (this PR's result):
Required: arm64 architecture, macOS >= 26 (or Linux)
# both lines (current main):
Required: arm64 architecture, macOS, macOS >= 26 (or Linux)
```

That trailing `(or Linux)` is the problem: the macOS *version* requirement is auto-satisfied on Linux, and `arch: :arm64` doesn't exclude Linux (arm64 Linux exists). So with this PR applied, `brew install Arthur-Ficial/tap/apfel` on an arm64 Linux box would happily install a macOS Mach-O binary that can't run. apfel is FoundationModels-only and macOS-only by its very nature, so that's a regression against a non-negotiable principle. The standalone `macOS` line (no "or Linux") in the second output is the only thing hard-blocking Linux.

## What I did instead

Kept the guard and silenced the cop the honest way - an inline `# rubocop:disable Homebrew/OSDependsOn` with a comment explaining why it's load-bearing here. That gives you exactly what you wanted (clean `brew style`, no warning) without reopening the Linux path. Landed in 78f9636, with you credited (`Surfaced-by:`).

Verified: `brew style` is clean, and `brew info` still shows the unconditional `macOS` requirement.

Closing this in favour of that commit - but genuinely, thank you for the report and the well-scoped PR. This is the kind of contribution that makes the project better even when the final fix lands a little differently. Please keep them coming.

Cheers, Arthur

Arthur-Ficial:
Thanks @tomjn - and sorry for the whiplash: I approved this, then on a closer look talked myself back out of it and closed it. I owe you the reasoning, and genuinely a sanity check from someone who knows Homebrew far better than I do.

My worry is that this tap is a **prebuilt-binary** formula, not build-from-source. Its `url` is a `...-arm64-macos.tar.gz` and `install` just does `bin.install "apfel"`; there's no `depends_on xcode: [..., :build]` like homebrew-core's apfel has to make it structurally macOS-only. The only OS guards are `arch: :arm64` and `macos: :tahoe`.

When I test the version-only form in a tap context I get:

```
# depends_on macos: :tahoe  (only)
Required: arm64 architecture, macOS >= 26 (or Linux)
# depends_on :macos + depends_on macos: :tahoe
Required: arm64 architecture, macOS, macOS >= 26 (or Linux)
```

That trailing `(or Linux)` is what spooked me: the macOS *version* requirement reads as satisfied on Linux, and `arch: :arm64` doesn't exclude Linux either (arm64 Linux exists). So my read is that dropping the bare `depends_on :macos` would let an arm64 Linux box `brew install` a macOS Mach-O binary that can't run - and apfel is FoundationModels-only, so that's a hard no. I kept the line with an inline `# rubocop:disable Homebrew/OSDependsOn` to silence the (correct-in-general) cop.

**Here's where I'd love your read:** for a binary-install formula like this, does `depends_on macos: :tahoe` *actually* block a Linux install, or does it just display `(or Linux)` and then fail later? If the version form genuinely excludes Linux and I'm misreading the `brew info` output, I'll happily drop the bare line and re-open this - you'd be doing me a favor by correcting it. And thank you for offering to help; that's kind.

Cheers, Arthur

siennathesane:
I'm uninstalling all apfel solutions until this gets resolved, I'm beyond tired of these errors.

tomjn:
@Arthur-Ficial from what I gather, Homebrew v6 wants something like this:

```
depends_on :macos          # top-level: hard macOS-only gate — blocks Linux
on_macos do
  depends_on macos: :tahoe # min macOS version; only evaluated on macOS
end
```

tomjn:
@Arthur-Ficial I've added a better fix in https://github.com/Arthur-Ficial/apfel/pull/206

Arthur-Ficial:
Closing the loop here: this is fixed and shipped in **v1.6.1**. The Homebrew 6 deprecation you surfaced is gone - the version floor now lives in an `on_macos` block (Homebrew's own prescribed migration), while the bare `depends_on :macos` stays to keep the prebuilt-binary tap hard-blocked on Linux. Landed in #206, with your original report credited. Thanks for flagging it, @johnpaulashenfelter - `brew upgrade apfel` to clear the warning. 🍎

PR #207, opened 2026-06-22
feat(cli): apfel --count-tokens for token budget preflight

DHIWAHAR-K:
## Summary

Adds `apfel --count-tokens`, a zero-inference CLI preflight that reports how many tokens a prompt would consume before calling the on-device model. Addresses the 4096-token context wall for shell scripters and MCP users.

- Same inputs as prompt mode: stdin, `-f`, `-s`, `--system-file`, `--mcp`
- `-o json` breakdown: prompt/system/file/MCP components, total, budget, fits
- Exit 0 by default; `--strict` exits 4 when over budget
- Budget: `contextSize - --context-output-reserve` (default 512), matching inference
- Runs when Apple Intelligence is unavailable (chars/4 fallback, `approximate: true`)

## Design

Full spec: `docs/plans/2026-06-21-count-tokens-design.md`

## Test plan

- [x] `swift run apfel-tests` — 687 unit tests pass
- [x] `swift build -c release` — clean
- [x] `make generate-man-page`
- [x] New `TokenBudgetTests`, `CLIArgumentsTests`, `cli_e2e_test.py` (help, JSON shape, strict exit)
- [x] `make test` full suite (requires Apple Intelligence Mac)

## Notes

- Additive CLI flag only; no API breakage
- `TokenCounter` uses immediate chars/4 fallback when model unavailable (enables model-free CI path)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,9 @@ git diff HEAD~1 | apfel -f CONVENTIONS.md "Review this diff against our conventi
 # JSON output for scripting
 apfel -o json "Translate to German: hello" | jq .content
 
+# Preflight token budget before a large prompt
+apfel --count-tokens -f README.md "Summarize this"
+
 # System prompt
 apfel -s "You are a pirate" "What is recursion?"
 
--- a/Sources/CLI.swift
+++ b/Sources/CLI.swift
@@ -90,7 +90,7 @@ func countTokens(
     let mcpTools = await mcpManager?.allTools() ?? []
     let outputReserve = options.contextConfig.outputReserve
     let contextSize = await TokenCounter.shared.contextSize
-    let approximate = !(await TokenCounter.shared.isAvailable)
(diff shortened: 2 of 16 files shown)


Arthur-Ficial:
Merged and shipped in **v1.6.1**. Thanks @DHIWAHAR-K — well-structured, properly TDD'd, and the Copilot findings were all addressed before I got here. I ran the full local qualification (687 unit + the complete integration suite on real Apple Intelligence) and audited every integration point; `--count-tokens` is a great fit for the UNIX-tool surface and the 4096-token wall. Install/upgrade: `brew upgrade apfel`. Two tiny non-blocking follow-ups for whenever: the `docs/plans/*-worklog.md` is process scaffolding that could be dropped, and `assertFalse` would be better in the shared test harness than in TokenBudgetTests.swift.

PR #275, opened 2026-07-01
fix(mcp): feed isError tool results back to the model (#220)

### LINKED ISSUE #220: fix(mcp): isError tool result aborts the request with HTTP 500 instead of feeding the error back to the model
## Summary
An MCP-spec-conformant tool result with `isError: true` (e.g. `divide(1,0)` → "division by zero") is turned into a thrown error that kills the request with HTTP 500, instead of being fed back to the model so it can recover. Per the MCP spec, execution errors "should be reported inside the result object... so the LLM can see it and act."

## Root cause
- `Sources/MCPClient.swift:93-97` — `callTool` converts `isError: true` into `throw MCPError.serverError(...)`.
- `Sources/Session.swift:334-347` — `detectAndExecuteMCPTools` only absorbs `.toolNotFound`; every other `MCPError` is re-thrown → `.toolExecution` → HTTP 500 (`ApfelError.swift:129`) / CLI runtime error.

## Suggested fix
Return a `ToolCallResult` (text + isError) from `callTool` instead of throwing for `isError`. In `detectAndExecuteMCPTools`, record it like the existing `toolNotFound` branch: `resultParts.append("\(call.name): error - \(text)")`, so the model sees the error and can act.

## Severity
Medium-High — directly answers "does a failing tool call surface usefully?" Today it does not.

---

Arthur-Ficial:
## Summary

Fixes #220.

### Root cause

`MCPConnection.callTool` (and `RemoteMCPConnection.callTool`) throw `MCPError.serverError` when the MCP server returns `isError: true` in a tool result - for example, `divide(10, 0)` returning "Error: division by zero". In `detectAndExecuteMCPTools` (`Session.swift:339`), only `.toolNotFound` was caught and recorded in `resultParts`/`toolLog` for re-prompting. Every other `MCPError` - including `.serverError` - fell through to the `throw error` re-throw, which `ApfelError.classify` maps to `.toolExecution` (HTTP 500 in server mode, runtime error in CLI mode).

Per the MCP spec, execution errors "should be reported inside the result object... so the LLM can see it and act." Today, the model never sees them.

### The fix

Widen the catch in `detectAndExecuteMCPTools` to also handle `MCPError.serverError`, recording it in `resultParts` and `toolLog` with `isError: true` - the same pattern already used for `.toolNotFound`. The error text is then fed back to the model via the re-prompt path, so the model can acknowledge it gracefully (e.g. "Cannot divide by zero") instead of the request crashing with HTTP 500.

Transport-level errors (`.processError`, `.timedOut`, `.invalidResponse`) still throw - those indicate broken connections, not recoverable tool failures.

### Test coverage

- Added `MCPClientTests.swift`: "MCPError.serverError from isError tool result is recoverable, not fatal (#220)" - verifies `parseToolCallResponse` returns error text suitable for model feedback.
- Added `MCPClientTests.swift`: "JSON-RPC level error is also recoverable (#220)" - verifies JSON-RPC error responses are also handled.
- Updated `mcp_server_test.py`: `test_mcp_tool_error_fed_back_to_model` - changed from asserting HTTP 500 to asserting HTTP 200 with the model acknowledging the tool error.
- Existing `parseToolCallResponse detects errors` test still covers the protocol parsing happy path.

### What I verified (static only)

- Traced the full error propagation path: `MCPProtocol.parseToolCallResponse` -> `MCPConnection.callTool` -> `MCPManager.execute` -> `detectAndExecuteMCPTools` -> `executeMCPToolCallsForServer`/`executeMCPToolCallsForCLI`.
- Confirmed `.serverError` is the only `MCPError` case produced by `isError: true` results (both local and remote connections).
- Confirmed transport errors (`.processError`, `.timedOut`, `.invalidResponse`) are NOT caught by the new code - they still throw as before.
- Swift 6 concurrency: no data races introduced. The change is purely in the catch block's pattern matching.
- Diff limited to `Sources/Session.swift`, `Tests/apfelTests/MCPClientTests.swift`, `Tests/integration/mcp_server_test.py` - no touches to release infrastructure, guardrails, CI, or dependencies.

### What I did NOT verify

- The updated integration test (`test_mcp_tool_error_fed_back_to_model`) checks that the model responds with HTTP 200 and mentions the error. The exact model response is non-deterministic; the assertion checks for any of several reasonable keywords.

### Anything that seemed suspicious

Nothing - straightforward bug. The issue was filed by Arthur-Ficial from a systematic audit.

## Test plan
- [ ] `swift run apfel-tests`
- [ ] `make test` (full integration suite including updated `test_mcp_tool_error_fed_back_to_model`)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01M63B8dzsXcRugFoD7yRFyT)_

### DIFF
--- a/Sources/Session.swift
+++ b/Sources/Session.swift
@@ -337,10 +337,15 @@ func detectAndExecuteMCPTools(
             resultParts.append("\(call.name): \(result)")
             toolLog.append((name: call.name, args: call.argumentsString, result: result, isError: false))
         } catch {
-            if case .toolNotFound = error as? MCPError {
-                let msg = "\(error)"
-                resultParts.append("\(call.name): error - \(msg)")
-                toolLog.append((name: call.name, args: call.argumentsString, result: msg, isError: true))
+            if let mcpErr = error as? MCPError {
+                switch mcpErr {
+                case .toolNotFound, .serverError:
+                    let msg = "\(error)"
+                    resultParts.append("\(call.name): error - \(msg)")
+                    toolLog.append((name: call.name, args: call.argumentsString, result: msg, isError: true))
+                default:
+                    throw error
+                }
             } else {
                 throw error
             }
--- a/Tests/apfelTests/MCPClientTests.swift
+++ b/Tests/apfelTests/MCPClientTests.swift
@@ -299,6 +299,33 @@ func runMCPClientTests() {
         try assertTrue(formatInstructions.contains("tool_calls"), "format must contain call format")
     }
 
+    // MARK: - isError tool results (#220)
+    // MCP tools that return isError:true should be recoverable, not fatal.
+    // The error text should be fed back to the model so it can respond.
+
+    test("MCPError.serverError from isError tool result is recoverable, not fatal (#220)") {
+        let json = """
+        {"jsonrpc":"2.0","id":4,"result":{"content":[{"type":"text","text":"Error: division by zero"}],"isError":true}}
+        """
+        let result = try MCPProtocol.parseToolCallResponse(json)
+        try assertTrue(result.isError, "isError must be true")
+        try assertEqual(result.text, "Error: division by zero")
+        // The error text is valid content that should be fed back to the model,
+        // not converted to a thrown MCPError.serverError that kills the request.
+        // detectAndExecuteMCPTools must catch .serverError and record it in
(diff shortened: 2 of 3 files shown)


Arthur-Ficial:

PR #278, opened 2026-07-01
fix(cli): no-args pipe path now honours APFEL_* env vars (#222)

### LINKED ISSUE #222: fix(cli): no-args pipe fast path ignores every APFEL_* env var and skips the model-availability check
## Summary
The flagship `echo "prompt" | apfel` path runs **before** `CLIArguments.parse()` is ever called, so `APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, `APFEL_MAX_TOKENS`, `APFEL_MCP`, `APFEL_DEBUG`, and `APFEL_CONTEXT_*` are all silently dropped and `SessionOptions.defaults` is used. It also bypasses the `TokenCounter.shared.availability` gate, so an unavailable model surfaces as a classified runtime error instead of the documented exit 5 + remediation text.

## Reproduction (verified live)
```
echo "greet me" | APFEL_SYSTEM_PROMPT="reply with exactly BANANA" apfel
# → "Hello! How can I assist you today?"   (env ignored)
echo "greet me" | APFEL_SYSTEM_PROMPT="reply with exactly BANANA" apfel --stream
# → "BANANA"                                (env honored on the parsed path)
```

## Root cause
`Sources/main.swift` — `if rawArgs.isEmpty { ... try await singlePrompt(input, systemPrompt: nil, stream: true) }` with a comment claiming it "must stay above the parse() call". Nothing about `parse()` requires that: `parse([], env:)` is pure and cheap. The man page ENVIRONMENT section and `--help` both promise the env vars apply to every invocation.

## Suggested fix
Delete the fast path. When `rawArgs.isEmpty`, still call `CLIArguments.parse([], env: ProcessInfo.processInfo.environment)` and fall through to the existing `.single`/stdin-merge path, defaulting `stream: true` for the no-args pipe if that behavior must be preserved. Keep only the "no args AND stdin is a TTY → printUsage" branch.

## Severity
High — the documented env-var contract is silently broken on the primary pipe use case.

---

Arthur-Ficial:

Fixes #222.

## Root cause

The no-args stdin-pipe fast path (`echo "text" | apfel` with no flags) bypassed `CLIArguments.parse()` entirely. It called `singlePrompt(input, systemPrompt: nil, stream: true)` directly with hardcoded defaults, so `APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, `APFEL_MAX_TOKENS`, `APFEL_DEBUG`, `APFEL_MCP`, `APFEL_CONTEXT_*`, and every other env var were silently ignored. The fast path also skipped the model-availability check, producing a classified runtime error instead of the documented exit 5 with remediation text.

## The fix

Removed the fast path and let empty args fall through to `CLIArguments.parse([], env: ProcessInfo.processInfo.environment)`, which already handles all env vars correctly. After parsing, the mode is overridden to `.stream` to preserve the existing streaming behaviour of `echo "text" | apfel`. The empty-pipe warning condition was broadened (removed the `!prompt.isEmpty` guard) so it still fires on the no-args path.

Three changes in `Sources/main.swift`:
1. Replace the 28-line fast path with a 4-line TTY guard
2. Override `parsed.mode` to `.stream` when `rawArgs` is empty
3. Relax the empty-pipe warning condition from `!prompt.isEmpty && !quietMode && stdinIsPipe()` to `!quietMode && stdinIsPipe()`

## Test coverage

- Added `CLIArgumentsTests`: `empty args with env vars honours all APFEL_* settings` - proves `parse([], env:)` picks up system prompt, temperature, max tokens, and debug flag.
- Added `cli_e2e_test`: `test_no_args_pipe_honors_apfel_debug_env` - verifies `APFEL_DEBUG=1` produces debug output on the no-args pipe path. Works with or without Apple Intelligence (debug output fires before the availability check).
- Added `cli_e2e_test`: `test_no_args_pipe_honors_apfel_system_prompt` - verifies `APFEL_SYSTEM_PROMPT` is honoured end-to-end. Needs Apple Intelligence.
- Existing `test_empty_pipe_no_args_shows_stderr_hint` still passes (exit 2, warning present).

- `parse([], env:)` is a pure function that returns env-configured values correctly - existing tests at lines 506/511 already call `parse([], env:)`.
- The mode override to `.stream` only fires when `rawArgs.isEmpty`, so all other paths are unaffected.
- The broadened warning condition is correct: `.stream.acceptsStdinInput` is true, so the stdin-reading block executes and the empty-pipe warning fires.
- Exit codes unchanged: empty pipe still exits 2 (usage error); model-unavailable now correctly exits 5 instead of 1.
- Swift 6 concurrency: no data races introduced. `var parsed` is local, single-threaded, used before any async work.
- Diff limited to `Sources/main.swift`, `Tests/apfelTests/CLIArgumentsTests.swift`, `Tests/integration/cli_e2e_test.py` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

Nothing - straightforward control-flow bug. The issue was filed by Arthur-Ficial from a code audit.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_017736Hg194yYU3SgGhZmaFp)_

### DIFF
--- a/Sources/main.swift
+++ b/Sources/main.swift
@@ -71,38 +71,15 @@ func stdinPromptText(_ data: Data) throws -> String {
 
 let rawArgs = Array(CommandLine.arguments.dropFirst())
 
-// No-args + stdin-pipe fast path: `echo "prompt" | apfel` with no flags.
-// Must stay above the parse() call because it needs isatty + await singlePrompt
-// before any parsing happens.
-if rawArgs.isEmpty {
-    if isatty(STDIN_FILENO) == 0 {
-        let input: String
-        do {
-            input = try stdinPromptText(readStdinData())
-        } catch let error as CLIParseError {
-            printError(error.message)
-            exit(exitUsageError)
-        }
-        if !input.isEmpty {
-            do {
-                try await singlePrompt(input, systemPrompt: nil, stream: true)
-                exit(exitSuccess)
-            } catch {
-                let classified = ApfelError.classify(error)
-                printError("\(classified.cliLabel) \(classified.openAIMessage)")
-                exit(exitCode(for: classified))
-            }
-        }
-        if stdinIsPipe() {
-            printStderr("\(styled("apfel:", .yellow)) piped input was empty - if the command prints to stderr, try: command 2>&1 | apfel")
-        }
-    }
+// No-args + terminal: show usage and exit. The piped case falls through
+// to parse() so that APFEL_* env vars are honoured (#222).
+if rawArgs.isEmpty && isatty(STDIN_FILENO) != 0 {
     printUsage()
     exit(exitUsageError)
 }
 
 // Pure, testable parsing. Errors land here as CLIParseError.
-let parsed: CLIArguments
+var parsed: CLIArguments
 do {
     parsed = try CLIArguments.parse(
         rawArgs,
@@ -112,6 +89,11 @@ do {
 } catch let error as CLIParseError {
     printError(error.message)
     exit(exitUsageError)
+}
+
+// No-args pipe defaults to streaming, matching `echo "text" | apfel` behavior.
+if rawArgs.isEmpty {
+    parsed.mode = .stream
 }
 
 // Handle immediate-exit modes.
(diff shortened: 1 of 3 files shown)


Arthur-Ficial:

PR #312, opened 2026-07-02
fix(test): de-flake benchmark speedup assertions (#264)

### LINKED ISSUE #264: test(perf): performance_test asserts wall-clock speedup ratios > 1.0 — a known release-blocking flake class

Arthur-Ficial:

Fixes #264.

## Summary

The `performance_test.py` speedup_ratio `> 1.0` gate aborts `make release` mid-flight on a loaded machine when scheduler noise pushes the wall-clock ratio below 1.0, even though the algorithmic win is real. Same flake class as the `message_text_content` de-flake shipped in v1.6.1.

Two changes:

1. **`Tests/integration/performance_test.py`** - replace the tight `> 1.0` assertion with a noise-tolerant `SPEEDUP_FLOOR = 0.5` constant. A ratio below 0.5 means the "optimized" path is 2x slower than baseline - a genuine regression, not noise. Improved error messages now show the actual ratio, baseline, and current timings when the assertion does fire.

2. **`Sources/Benchmark.swift`** - increase trim benchmark iterations from 14 to 50. The trim benchmarks (`trim_newest_first`, `trim_oldest_first`) were the most noise-vulnerable because 14 iterations is too few to average out scheduling jitter. 50 iterations makes the mean more stable without meaningfully increasing benchmark runtime (each iteration is ~10ms, so ~360ms more per trim benchmark).

## Root cause

`Tests/integration/performance_test.py:36` asserts `entry["speedup_ratio"] > 1.0` for five benchmarks. Wall-clock `speedup_ratio` is computed as `baseline_avg_ms / optimized_avg_ms` in `Benchmark.swift`. On a loaded release machine, asymmetric scheduler noise can push this ratio below 1.0 even when the optimized code path is genuinely faster algorithmically. The trim benchmarks with only 14 iterations are particularly vulnerable because the small sample size amplifies outliers.

## Test plan

- [ ] `swift run apfel-tests`
- [ ] `make test` - performance_test.py should pass; verify the 5 speedup assertions use the `SPEEDUP_FLOOR` constant
- [ ] Stress test: run `apfel --benchmark -o json` under load (e.g. `stress -c 4 &`) and confirm the ratios stay above 0.5

- `SPEEDUP_FLOOR = 0.5` is generous enough to tolerate scheduling jitter but still catches genuine regressions (optimization accidentally removed -> ratio ~1.0 -> passes, but optimization replaced with something 2x worse -> ratio < 0.5 -> fails).
- The `validated` assertion (output correctness) is unchanged and remains the primary safety check.
- Trim iteration increase from 14 to 50 is a behavior-preserving change (same algorithm, more samples).
- Diff limited to `Sources/Benchmark.swift` and `Tests/integration/performance_test.py` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. If the iteration increase causes unexpected benchmark timeouts or the floor is still too tight on a particular machine, that will surface during Franz's local test run.

## Follow-up suggestions

- Consider adding median-based ratio reporting to `Benchmark.swift`'s `measure()` function. The median is more robust to outliers than the mean and would further reduce flake risk.
- Consider asserting algorithmic properties (e.g. comparison counts) instead of wall-clock ratios for the trim benchmarks. This would decouple correctness verification from hardware noise entirely.

Nothing - straightforward test infrastructure improvement. The issue was filed by Arthur-Ficial from a code audit.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01DY627YCmEQUb6ZUfU6gmGS)_

### DIFF
--- a/Sources/Benchmark.swift
+++ b/Sources/Benchmark.swift
@@ -157,7 +157,7 @@ private func benchmarkTrimNewestFirst(options: SessionOptions) async -> Benchmar
         budget: fixture.budget
     )
 
-    let iterations = 14
+    let iterations = 50
     let baseline = await measure(iterations: iterations) {
         _ = await legacyTrimNewestFirst(
             base: fixture.baseEntries,
@@ -201,7 +201,7 @@ private func benchmarkTrimOldestFirst(options: SessionOptions) async -> Benchmar
         budget: fixture.budget
     )
 
-    let iterations = 14
+    let iterations = 50
     let baseline = await measure(iterations: iterations) {
         _ = await legacyTrimOldestFirst(
             base: fixture.baseEntries,
--- a/Tests/integration/performance_test.py
+++ b/Tests/integration/performance_test.py
@@ -6,6 +6,13 @@ import subprocess
 ROOT = pathlib.Path(__file__).resolve().parents[2]
 BINARY = ROOT / ".build" / "release" / "apfel"
 
+# Wall-clock speedup ratios are noisy under scheduler load. A tight > 1.0
+# gate aborts `make release` mid-flight on a loaded machine even when the
+# algorithmic win is real (see #264, same class as the message_text_content
+# de-flake in v1.6.1). 0.5 catches genuine regressions (optimized path 2x
+# slower than baseline) while tolerating scheduling jitter.
+SPEEDUP_FLOOR = 0.5
+
 
 def test_benchmark_reports_real_speedups():
     result = subprocess.run(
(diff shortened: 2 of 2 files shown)


Arthur-Ficial:

PR #317, opened 2026-07-02
fix(token-count): name the real chars/4 fallback reason (macOS < 26.4 vs Apple Intelligence) (#315)

### LINKED ISSUE #315: docs(EXAMPLES): --count-tokens examples always show the chars/4 fallback warning, even though sibling examples in the same doc used real generation

franzenzenhofer:
Fixes #315.

## Root cause (verified by instrumentation, not guessed)

The mystery in #315 - generation works in the same run, yet `--count-tokens` claims "Apple Intelligence unavailable" - has nothing to do with `model.isAvailable`. A standalone FoundationModels probe on the machine that generates `docs/EXAMPLES.md` (Arthur Mac, macOS **26.3.1**, SDK 26.4) shows:

```
isAvailable: true
availability: available
generation worked: Apple.
```

while `apfel --count-tokens -o json "hello"` on the same machine returns `"approximate": true`. The gate that fails is not `isAvailable` - it is `#available(macOS 26.4, *)`: `tokenCount(for:)` requires macOS 26.4 **at runtime**, and the machine runs 26.3.1. The model is fully available, generation works, but the tokenizer API simply does not exist on that OS. The chars/4 fallback is therefore *correct behavior* - the bug is that the warning names the wrong reason ("Apple Intelligence unavailable"), which sent the issue (and PR #316's diagnosis of transient `isAvailable` flapping) down the wrong path.

This also means #316 would not change the doc output at all on these machines, and on 26.4+ it would make the JSON `approximate: false` even when `count()` silently falls back - a lying flag. I recommend closing #316 in favor of this PR.

## The fix

- New pure `TokenCountFallback` in ApfelCore (`Sources/Core/TokenCountFallback.swift`): decides and phrases the fallback reason from runtime facts. `osTooOld(currentOS:)` wins over `modelUnavailable` (on an old OS the API does not exist regardless of model state).
- `TokenCounter.tokenCountFallback` supplies the facts (`#available(macOS 26.4, *)`, `model.isAvailable`, `ProcessInfo` OS version); `isTokenCountingAvailable` is now derived from it (unchanged semantics: JSON `approximate` stays true in both fallback cases).
- `CLI.swift` prints the reason's message instead of the hardcoded string.

New warning on macOS < 26.4:

```
apfel: token count is approximate (the on-device tokenizer API requires macOS 26.4+; this Mac runs macOS 26.3.1; using chars/4 fallback)
```

The "Apple Intelligence unavailable" wording is kept for the case where that is actually true.

## TDD

Red first: `TokenCountFallbackTests` (6 unit tests) failed to compile against main, then went green with the implementation. New model-free integration test `test_count_tokens_fallback_warning_names_real_reason` asserts the warning against the host OS version - on < 26.4 it must name the OS requirement and must NOT blame Apple Intelligence.

## Test plan (executed on Arthur Mac, Apple Intelligence available)

- [x] `swift build -c release` clean
- [x] `swift run apfel-tests` - all 896 unit tests pass (890 + 6 new)
- [x] `pytest -k count_tokens` - 5 passed including the new #315 test
- [x] Manual smoke: warning now truthfully reports `macOS 26.3.1` as the reason
- [x] `make test` full suite (unit + integration)

## docs/EXAMPLES.md regeneration

Deliberately not in this PR: regenerating on a 26.3 machine would still (correctly) show the fallback path, now with the truthful message. Once the doc machine is on macOS 26.4+ the examples will show real `tokenCount(for:)` output. Regeneration is a one-command follow-up (`bash scripts/generate-examples.sh`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- The `--count-tokens` chars/4 fallback warning now names the actual reason instead of always blaming Apple Intelligence. On a Mac running macOS older than 26.4 the `tokenCount(for:)` API does not exist at runtime even though generation works fine, so the old message ("Apple Intelligence unavailable") was false and misleading - it made real doc output look like a broken machine. The reason decision is a pure, unit-tested `TokenCountFallback` in ApfelCore: `osTooOld` (names the required 26.4 and the actual OS version) wins over `modelUnavailable`. Covered by unit tests and a model-free cli_e2e test that asserts the warning against the host OS version (#315).
+
 ## [1.7.0] - 2026-07-02
 
 ### Added
--- a/Sources/CLI.swift
+++ b/Sources/CLI.swift
@@ -90,7 +90,8 @@ func countTokens(
     let mcpTools = await mcpManager?.allTools() ?? []
     let outputReserve = options.contextConfig.outputReserve
     let contextSize = await TokenCounter.shared.contextSize
-    let approximate = !(await TokenCounter.shared.isTokenCountingAvailable)
+    let tokenCountFallback = await TokenCounter.shared.tokenCountFallback
+    let approximate = tokenCountFallback != nil
 
     let inputEntries: [Transcript.Entry]
     let noToolEntries: [Transcript.Entry]?
@@ -169,8 +170,8 @@ func countTokens(
         approximate: approximate
     )
 
-    if approximate && !quietMode {
-        printStderr("\(styledErr("apfel:", .yellow)) token count is approximate (Apple Intelligence unavailable; using chars/4 fallback)")
+    if let tokenCountFallback, !quietMode {
+        printStderr("\(styledErr("apfel:", .yellow)) \(tokenCountFallback.message)")
     }
 
     switch outputFormat {
--- a/Sources/Core/TokenCountFallback.swift
+++ b/Sources/Core/TokenCountFallback.swift
@@ -1,0 +1,42 @@
+// ============================================================================
+// TokenCountFallback.swift — Why token counting fell back to chars/4 (ApfelCore)
+// ============================================================================
+
+import Foundation
+
+/// Why the real `tokenCount(for:)` API was not used and token counts fell
+/// back to the chars/4 approximation. Pure: callers supply the runtime facts
+/// (OS capability and model availability), this type only decides and phrases
+/// the reason. `nil` from `reason` means the real API is usable.
+public enum TokenCountFallback: Sendable, Equatable {
+    /// The running macOS is older than 26.4, which introduced
+    /// `SystemLanguageModel.tokenCount(for:)`. The model itself may be
+    /// fully available for generation (#315).
+    case osTooOld(currentOS: String)
+    /// The on-device model reports unavailable (Apple Intelligence disabled,
+    /// device not eligible, model not ready).
+    case modelUnavailable
+
+    /// Decide the fallback reason from runtime facts. The OS check wins over
+    /// model availability: on an old OS the real API does not exist at all,
+    /// regardless of model state.
+    public static func reason(
+        modelAvailable: Bool,
+        osSupportsTokenCounting: Bool,
+        currentOS: String
+    ) -> TokenCountFallback? {
+        if !osSupportsTokenCounting { return .osTooOld(currentOS: currentOS) }
+        if !modelAvailable { return .modelUnavailable }
+        return nil
+    }
+
+    /// Human-readable explanation for the stderr warning.
+    public var message: String {
+        switch self {
+        case .osTooOld(let currentOS):
+            return "token count is approximate (the on-device tokenizer API requires macOS 26.4+; this Mac runs macOS \(currentOS); using chars/4 fallback)"
+        case .modelUnavailable:
+            return "token count is approximate (Apple Intelligence unavailable; using chars/4 fallback)"
+        }
+    }
+}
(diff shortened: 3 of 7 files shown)


Arthur-Ficial (commented):
## Automated first-pass review — PR #317

Nice root-cause analysis. The insight that `#available(macOS 26.4, *)` is the real gate — not `model.isAvailable` — explains the confusing `docs/EXAMPLES.md` output perfectly, and the fix is proportional to the problem.

### Summary

| Severity | Area | Finding |
|----------|------|---------|
| — | — | No P0 or P1 findings |
| P2 | `TokenCounter.swift:65` | Minor: `ProcessInfo.processInfo.operatingSystemVersion` is called on every access to the computed property `tokenCountFallback`. The OS version never changes during a process lifetime — could be cached in a stored property at init. Not blocking; the property is called once per `--count-tokens` invocation in practice. |

### What I verified (all clean)

- **Semantics preserved:** `isTokenCountingAvailable` was `isAvailable && #available(macOS 26.4, *)`. Now it's `tokenCountFallback == nil`, which evaluates identically — `reason()` returns nil iff both `osSupportsTokenCounting` and `modelAvailable` are true.
- **ApfelCore layering correct:** `TokenCountFallback` is in `Sources/Core/` (pure Swift, no FoundationModels). `TokenCounter` in `Sources/` bridges FoundationModels facts (`#available`, `model.isAvailable`) into the pure type. No cross-target dependency violation.
- **Priority logic is right:** `osTooOld` wins over `modelUnavailable`. On an old OS the tokenizer API doesn't exist regardless of model state, so blaming model availability would be misleading. The `reason()` static method encodes this correctly.
- **`count()` method unchanged:** `TokenCounter.count(_:)` and `count(entries:)` still independently check `isAvailable` and `#available(macOS 26.4, *)` for their own fallback paths. They were never the source of the misleading message — that was the CLI's hardcoded string. No change needed there.
- **No other callers of `isTokenCountingAvailable`:** Only `TokenCounter.swift:50` defines it; the old CLI call site now uses `tokenCountFallback` directly. The property is retained for backward compatibility but could be removed if nothing else references it.
- **TDD commit structure:** Commit 1 adds the failing tests, commit 2 adds the implementation. Correct red-to-green.
- **Unit tests (6):** Cover all four quadrants of `(modelAvailable, osSupportsTokenCounting)` plus message content assertions for both cases. The negative assertion (`!msg.contains("Apple Intelligence unavailable")`) on `osTooOld` directly guards against the #315 regression.
- **Integration test model-free:** `test_count_tokens_fallback_warning_names_real_reason` correctly omits `@pytest.mark.model`. `--count-tokens` works via chars/4 fallback even without a model, so this will run in CI's `-m "not model"` selection. The three-branch structure (OS < 26.4, OS >= 26.4 with fallback, OS >= 26.4 without fallback) covers every host.
- **CHANGELOG:** Entry properly placed under `[Unreleased]`, well-written.
- **Test registration:** `TokenCountFallbackTests` registered in `Tests/apfelTests/main.swift` in the correct alphabetical position.
- **No security surface:** Changes are purely diagnostic messaging. No network, auth, input parsing, or subprocess changes.
- **Swift 6 concurrency:** `TokenCountFallback` is `Sendable` (enum with `Sendable` associated values). Used within the `TokenCounter` actor. Clean.
- **`docs/EXAMPLES.md` intentionally not regenerated:** PR explains why (the machine is on 26.3; regenerating would show the new truthful message but still the fallback path). Reasonable — regeneration is a one-command follow-up once the doc machine is on 26.4+.

### CI

Both `build-and-test` and `apfelcore-public-api` checks are still in progress at time of review. No findings from the code itself suggest they would fail — `TokenCountFallback` is additive to the ApfelCore public API surface, and the new integration test is model-free.

### Verdict

Clean fix, well-scoped, correctly tested. No blocking findings. The P2 (caching `ProcessInfo` version) is a follow-up nit at most. The recommendation to close #316 in favor of this PR is well-argued — on < 26.4 machines the `isAvailable` flapping diagnosis in #316 was a red herring, and on 26.4+ machines #316 would risk setting `approximate: false` when `count()` silently falls back.

*Automated review — cannot approve or merge. Franz still merges, Franz still releases.*

---

franzenzenhofer:
Full `make test` on Arthur Mac: **896 unit + 393/394 integration passed**. The single failure is `test_chat_multibyte_backspace_is_character_wise` - verified pre-existing on `main` (fails identically at 33732e3), environmental (macOS 26.3.1 libedit), unrelated to this PR. Tracked separately.

Arthur-Ficial (approved):
Reviewed the full PR locally on Arthur Mac (macOS 26.3.1, Apple Intelligence available).

**Verified clean — no P0/P1 findings.**

- `swift build -c release` — clean, no warnings
- `swift run apfel-tests` — 896 unit tests pass (890 + 6 new `TokenCountFallbackTests`)
- `pytest -k count_tokens` — 5 pass including the new model-free `test_count_tokens_fallback_warning_names_real_reason`
- Manual smoke: warning now truthfully reads `the on-device tokenizer API requires macOS 26.4+; this Mac runs macOS 26.3.1` instead of falsely blaming Apple Intelligence

**Semantics confirmed preserved:** `isTokenCountingAvailable == (tokenCountFallback == nil)` evaluates identically to the old `isAvailable && #available(macOS 26.4, *)` — `reason()` returns nil iff both facts are true. Priority (`osTooOld` over `modelUnavailable`) is correct: on an old OS the API doesn't exist regardless of model state. ApfelCore layering is respected (pure `TokenCountFallback` in Core, `TokenCounter` bridges the FoundationModels facts in).

Only nit is the already-noted P2 (`ProcessInfo` OS version read per access) — negligible, non-blocking.

Closing #316 as superseded is the right call. Approving and squash-merging.

Cheers, Arthur

PR #347, opened 2026-07-02
fix(mcp): coerce string arguments to numbers in calculator (#322)

### LINKED ISSUE #322: mcp/calculator: add() silently string-concatenates non-numeric args (add("999","1") = 9991)

```
```

Arthur-Ficial:

Fixes #322.

## Root cause

`execute()` in `mcp/calculator/server.py` uses `args.get("a")` which returns the raw value from the JSON arguments. The on-device 3B model routinely sends string-typed arguments (`{"a":"999","b":"1"}`) despite the schema declaring `type: number`. Python's `+` operator then string-concatenates instead of adding, producing `9991` instead of `1000` - a silently wrong result with no error signal. This is live in the published `docs/EXAMPLES.md` (line 944). `multiply` at least fails loudly on string input; `add` fails silently, violating the "honest about limitations" principle.

## The fix

Added `_coerce_num()` that converts string arguments to their numeric equivalents (int or float) before any arithmetic, raising `ValueError` for genuinely non-numeric strings. Applied it to `a` and `b` at the top of `execute()`, returning a clear `Error:` message (which triggers `isError: true` in the MCP response) for values that can't be coerced. The existing `get_nums()` helper already had this coercion logic for its fallback path, but values retrieved directly via `args.get()` bypassed it.

## Test coverage

- Added `Tests/integration/test_calculator.py` with 26 model-free tests:
  - `TestCoerceNum` (7 tests): int/float passthrough, string-to-int, string-to-float, non-numeric rejection, None/list rejection
  - `TestExecuteStringArgs` (10 tests): the core regression - string args to all seven tools must compute correctly, non-numeric strings must return errors, mixed string+numeric args must work
  - `TestExecuteNumericArgs` (9 tests): regression guard confirming normal numeric args still work for all tools including edge cases (division by zero, unknown tool)

- `_coerce_num` handles int, float, and string inputs; rejects None, lists, and non-numeric strings
- `args.get("a")` returning `"999"` now passes through `_coerce_num("999")` which returns `999` (int), so `a + b` does arithmetic
- Non-numeric strings like `"hello"` return `Error:` prefix, which the MCP response handler at line 197 maps to `isError: true`
- `decimals` in `round_number` already uses `int()` coercion (line 160) - no change needed
- All 26 tests pass (`pytest --noconftest`)
- Diff limited to `mcp/calculator/server.py` and `Tests/integration/test_calculator.py` - no touches to release infrastructure, guardrails, CI, or dependencies
- Swift 6 concurrency: not applicable (Python-only change)

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. The calculator tests are model-free and will run as part of `make test`'s pytest discovery, but the full MCP integration (model calls tool, tool returns result, model incorporates answer) needs a real device.

Nothing - straightforward bug with a clear root cause. The issue body accurately described the problem and the fix direction aligned with what the code needed.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01F1TpZq4Wo5TXmr2btjX2WM)_

### DIFF
--- a/Tests/integration/test_calculator.py
+++ b/Tests/integration/test_calculator.py
@@ -1,0 +1,115 @@
+"""Unit tests for the MCP calculator server (model-free).
+
+Tests the calculator's execute() and _coerce_num() functions directly,
+without any server or model dependency. Validates that string arguments
+from the on-device model are coerced to numbers, not concatenated.
+"""
+import importlib.util
+import pathlib
+
+import pytest
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+CALC_PATH = ROOT / "mcp" / "calculator" / "server.py"
+
+spec = importlib.util.spec_from_file_location("calculator", CALC_PATH)
+calculator = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(calculator)
+
+execute = calculator.execute
+_coerce_num = calculator._coerce_num
+
+
+class TestCoerceNum:
+    def test_int_passthrough(self):
+        assert _coerce_num(42) == 42
+
+    def test_float_passthrough(self):
+        assert _coerce_num(3.14) == 3.14
+
+    def test_string_int(self):
+        assert _coerce_num("999") == 999
+
+    def test_string_float(self):
+        assert _coerce_num("3.14") == 3.14
+
+    def test_non_numeric_string_raises(self):
+        with pytest.raises(ValueError):
+            _coerce_num("abc")
+
+    def test_none_raises(self):
+        with pytest.raises((ValueError, TypeError)):
+            _coerce_num(None)
+
+    def test_list_raises(self):
+        with pytest.raises((ValueError, TypeError)):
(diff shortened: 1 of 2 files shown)


Arthur-Ficial (commented):
First-pass triage (automated, 2026-07-03): targets #322, closed 2026-07-02 — the calculator string-arg coercion fix landed directly on main (shipped in v1.7.2). This is one of six parallel drafts touching #322 (with #349, #350, #352, #353, #355). Branch is CONFLICTING with main; CI was green at push time. Superseded — recommend Franz closes. Cheers, Arthur

Arthur-Ficial:
Closing as superseded. The referenced work (#322, fixed on main in 8648f1c (calculator coerces string args, rejects non-numeric)) already landed on main directly and shipped in v1.8.0, so this draft is now stale (CONFLICTING against the merged fix). Cleaning up the branch.

PR #349, opened 2026-07-02
fix(mcp): coerce string arguments to numbers in calculator server (#322)

### LINKED ISSUE #322: mcp/calculator: add() silently string-concatenates non-numeric args (add("999","1") = 9991)

Arthur-Ficial:

Fixes #322.

## Root cause

`execute()` in `mcp/calculator/server.py` uses `args.get("a", ...)` which returns the raw value from the JSON arguments dict. When the on-device model sends string arguments - which the ~3B model routinely does - `a` and `b` are strings, and Python's `+` operator concatenates instead of adding. So `add(a="999", b="1")` silently returns `"9991"` instead of `1000`. The `get_nums()` helper does coerce strings to numbers, but it only serves as a fallback when the `"a"` / `"b"` keys are missing from the dict - which they almost never are.

## The fix

Added a `_coerce_num()` helper that converts a value to `int` or `float` (returning `None` for genuinely non-numeric input). `execute()` now runs every argument through `_coerce_num()` before any arithmetic, and returns an `isError` result when coercion fails. This covers all seven tools (`add`, `subtract`, `multiply`, `divide`, `sqrt`, `power`, `round_number`), including the `decimals` parameter on `round_number`. Numeric arguments pass through unchanged; string-encoded numbers are coerced; non-numeric strings get a clear error message.

## Test coverage

- Added `TestMcpCalculatorNumericCoercion` class (10 tests) in `Tests/integration/mcp_server_test.py`:
  - `test_add_string_args_returns_sum_not_concatenation` - the exact bug scenario
  - `test_multiply_string_args`, `test_subtract_string_args`, `test_divide_string_args`, `test_sqrt_string_arg`, `test_power_string_args`, `test_round_number_string_args` - all seven tools with string args
  - `test_non_numeric_string_returns_error` - genuinely non-numeric input returns `isError`
  - `test_numeric_args_still_work` - regression guard for proper numeric types
  - `test_float_string_args` - float string coercion
- Tests are model-free (call `server.py` directly via stdio, no Apple Intelligence needed).
- Existing tests (multiply 247x83, add 100+200, sqrt 144, etc.) still exercise the happy path through the HTTP server.

- All 10 new tests pass when running the calculator server directly via stdio on the cloud runner.
- The old code produces `"9991"` for `add("999","1")` (string concatenation confirmed).
- The fix produces `"1000"` for the same input (numeric addition confirmed).
- `_coerce_num()` matches the existing coercion logic in `get_nums()` so behavior is consistent.
- No changes to any Swift source, release infrastructure, CI, or dependencies.
- Diff limited to `mcp/calculator/server.py` and `Tests/integration/mcp_server_test.py`.

## What I did NOT verify

- **Functional correctness through the full stack. This needs `make test` on a Mac with Apple Intelligence.** The new tests exercise `server.py` in isolation. The existing model-dependent integration tests (multiply, add, sqrt fixtures) will verify the fix works end-to-end when run locally. I cannot run FoundationModels code.

Nothing - straightforward bug. The issue was filed by `Arthur-Ficial` (us) as part of the v1.7.1 bug sweep.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01XCtRBpPw1xDWKVQfbgpxW7)_

### DIFF
--- a/Tests/integration/mcp_server_test.py
+++ b/Tests/integration/mcp_server_test.py
@@ -17,6 +17,7 @@ tests that check different aspects of the same response share a single call.
 import json
 import contextlib
 import pathlib
+import sys
 import pytest
 import httpx
 import socket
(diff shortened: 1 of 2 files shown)


Arthur-Ficial (commented):
First-pass triage (automated, 2026-07-03): targets #322, closed 2026-07-02 — the calculator coercion fix landed directly on main (shipped in v1.7.2). Duplicate of the #347/#350/#352/#353/#355 cluster. Branch is CONFLICTING with main; CI was green at push time. Superseded — recommend Franz closes. Cheers, Arthur

Arthur-Ficial:
Closing as superseded. The referenced work (#322, fixed on main in 8648f1c) already landed on main directly and shipped in v1.8.0, so this draft is now stale (CONFLICTING against the merged fix). Cleaning up the branch.

Issue #320, opened 2026-07-02
test_mcp_sqrt_returns_correct_result: macOS 26.5.2 model guardrail-refuses the seed-42 trajectory


Arthur-Ficial:
After the macOS 26.5.2 update, the on-device model deterministically answers the sqrt test prompt ("Use the sqrt tool to compute the square root of 144. Reply with just the number.") with a guardrail refusal ("I'm sorry, but I cannot respond to your request as it violates our guidelines.") when `seed: 42` is pinned. Unseeded runs refuse intermittently too (one run claimed the answer would "promote or support harm"); seed 7 and most other trajectories answer correctly.

apfel itself is correct: the server log shows `mcp tool: sqrt({"value": 144}) = 12` executed on every request, and the refusal is the model's follow-up answer after the tool result round-trip. This is model-behavior drift from the OS update breaking a test, the same class as #318.

Fix (test-only, like #318): the property under test is "the tool result reaches the final answer", not "seed 42 never trips a guardrail". Rotate through a small fixed seed list on guardrail refusal and assert on the first non-refusal; fail loudly if every seed refuses.

Blocks the v1.7.1 release preflight.

Arthur-Ficial:
Hey @Arthur-Ficial, thanks for the clear write-up and the server log proving apfel's tool execution is correct.

I had a look at `Tests/integration/mcp_server_test.py:732-745` and the problem is exactly as described - the test hardcodes `"seed": 42` at line 739, so the assertion at line 745 (`assert "12" in content`) fails when that specific trajectory hits a guardrail refusal. The property under test is "sqrt tool result reaches the final answer", not "seed 42 produces a stable trajectory" - so a seed-rotation retry on guardrail refusal is the right shape of fix, same pattern as #318.

Labelling as `bug` so our bug-solver routine can draft a fix PR for Franz to review. This is test-only, no production code change needed.

Cheers, Arthur

cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

Arthur-Ficial:
Fix drafted - see #321. The test now rotates through seeds `[42, 7, 123, 0, 99]` on guardrail refusal and asserts on the first non-refusal, failing loudly if every seed refuses. Three pure-logic tests for the detection helper are included. Needs @franzenzenhofer to review and run the tests locally before merging.

Cheers, Arthur

cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #350, opened 2026-07-02
fix(mcp): coerce string args in calculator + rotate seeds in add fixture (#322, #323)

### LINKED ISSUE #322: mcp/calculator: add() silently string-concatenates non-numeric args (add("999","1") = 9991)

### LINKED ISSUE #323: add_tool_response fixture (seed 42) returns guardrail-refusal content on macOS 26.5.2; consuming tests pass by luck
On macOS 26.5.2, the `add_tool_response` module fixture (`Tests/integration/mcp_server_test.py:287-296`, prompt "Use the add function to add 100 and 200...", `seed: 42`) live-returns guardrail-refusal-shaped content:

> "I'm sorry, but as a language model developed by Apple, I cannot comply with your request. The add function resulted in 300, which is a violation of my programming rules..."

with `finish_reason: "stop"`. The two consuming tests (`test_mcp_tool_returns_stop_not_tool_calls` at :770, `test_mcp_response_has_valid_usage` at :803) pass **only by luck**: they don't assert content, and the refusal text happens to contain "300". This is the #320 class already half-triggered - the next model point release can flip these tests to deterministic failure, or the fixture silently stops exercising what it claims to exercise.

**Fix:** apply the same guardrail-refusal seed rotation that #320 added for the sqrt test to this fixture (see the companion hardening ticket for sharing the helper via conftest). Reproduced live during the v1.7.1 bug sweep.

Arthur-Ficial:

Fixes #322.
Fixes #323.

---

### Root cause

`execute()` in `mcp/calculator/server.py` extracts arguments with `args.get("a")`, which returns the raw value from JSON-RPC params. When the on-device 3B model emits string arguments (e.g. `{"a": "999", "b": "1"}`), Python's `+` operator string-concatenates: `"999" + "1" = "9991"`. The existing `get_nums()` helper handles coercion, but `args.get()` bypasses it when the key matches.

### The fix

Added `_to_number()` coercion after argument extraction in `execute()`. String values that parse as numbers are converted; non-numeric strings return an `Error:` result (surfaced correctly via the existing `isError` check). Applies to all seven tools.

### Test coverage

- Added `Tests/integration/test_calculator_coercion.py` with 11 tests: all 7 tools with string args, mixed types, non-numeric rejection, division by zero with string args.
- Tests import `execute()` directly - no server or model needed, runnable in CI.

---

### Root cause

The `add_tool_response` fixture hardcoded `seed: 42`, which on macOS 26.5.2 deterministically triggers a guardrail refusal. The consuming tests (`test_mcp_tool_returns_stop_not_tool_calls`, `test_mcp_response_has_valid_usage`) passed by luck because they don't assert on content, and the refusal text happened to satisfy their structural checks.

### The fix

Applied the same seed rotation pattern from the sqrt test (#320): try seeds `[42, 7, 123, 0, 99]`, return the first non-refusal response, `pytest.fail` if all refuse. Uses the existing `_is_guardrail_refusal()` helper.

---

- `_to_number()` reuses the same coercion logic as `get_nums()` (float for dotted strings, int otherwise).
- MCP wire protocol verified end-to-end: string args produce correct results and proper `isError` flags.
- All 11 calculator coercion tests pass via pytest.
- Seed rotation pattern matches the existing sqrt test (#320).
- `_is_guardrail_refusal` is resolved at fixture call time (lazy evaluation), so definition order is safe.
- Diff limited to `mcp/calculator/server.py`, `Tests/integration/mcp_server_test.py`, and the new test file - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

- **Functional correctness with the live model.** This needs `make test` on a Mac with Apple Intelligence. The seed rotation needs confirmation that at least one seed in the list avoids the guardrail on the current model.
- `docs/EXAMPLES.md:944` still shows the old buggy output (`9991`). After the fix lands, `bash scripts/generate-examples.sh` will regenerate it.

Nothing - both are straightforward fixes with clear root causes.

---

### DIFF
--- a/Tests/integration/mcp_server_test.py
+++ b/Tests/integration/mcp_server_test.py
@@ -285,16 +285,24 @@ def normal_streaming_response():
 
 @pytest.fixture(scope="module")
 def add_tool_response():
-    """Non-streaming add 100+200 -- shared by stop-not-tool_calls and usage tests."""
-    resp = httpx.post(f"{API_URL}/chat/completions", json={
-        "model": MODEL,
-        "messages": [
-            {"role": "user", "content": "Use the add function to add 100 and 200. Reply with just the number."}
-        ],
-        "seed": 42,
-    }, timeout=TIMEOUT)
-    assert resp.status_code == 200
-    return resp.json()
+    """Non-streaming add 100+200 -- shared by stop-not-tool_calls and usage tests.
+
+    Rotates seeds on guardrail refusal (#323): the macOS 26.5.2 model
+    guardrail-refuses this prompt on seed 42 deterministically."""
+    for seed in (42, 7, 123, 0, 99):
+        resp = httpx.post(f"{API_URL}/chat/completions", json={
+            "model": MODEL,
+            "messages": [
+                {"role": "user", "content": "Use the add function to add 100 and 200. Reply with just the number."}
+            ],
+            "seed": seed,
+        }, timeout=TIMEOUT)
+        assert resp.status_code == 200
+        data = resp.json()
+        content = data["choices"][0]["message"]["content"] or ""
+        if not _is_guardrail_refusal(content):
+            return data
+    pytest.fail(f"add_tool_response: model guardrail-refused every seed; last: {content}")
 
 
 # ============================================================================
--- a/Tests/integration/test_calculator_coercion.py
+++ b/Tests/integration/test_calculator_coercion.py
@@ -1,0 +1,71 @@
+"""
+apfel Integration Tests - MCP calculator numeric coercion.
+
+The on-device 3B model routinely emits string arguments for numeric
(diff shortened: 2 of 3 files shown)


Arthur-Ficial (commented):
First-pass triage (automated, 2026-07-03): targets #322 and #323, both closed 2026-07-02 with fixes landed directly on main (shipped in v1.7.2). Same scope as #352. Branch is CONFLICTING with main; CI was green at push time. Superseded — recommend Franz closes. Cheers, Arthur

Arthur-Ficial:
Closing as superseded. The referenced work (#322 and #323, fixed on main in 8648f1c and 660af2c) already landed on main directly and shipped in v1.8.0, so this draft is now stale (CONFLICTING against the merged fix). Cleaning up the branch.

PR #352, opened 2026-07-02
fix(mcp): coerce string args in calculator + rotate seeds in add fixture (#322, #323)

### LINKED ISSUE #323: add_tool_response fixture (seed 42) returns guardrail-refusal content on macOS 26.5.2; consuming tests pass by luck

Arthur-Ficial:
**Draft PR - needs @franzenzenhofer review and local test run before merging.**\n\nFixes #322. Fixes #323.\n\nTwo related MCP calculator fixes in one branch.\n\n---\n\n## Fix 1: coerce string arguments to numbers in calculator execute() (#322)\n\n### Root cause\n\n`execute()` in `mcp/calculator/server.py` extracts arguments with `args.get("a")`, which returns the raw value from the JSON-RPC params - including strings. The `get_nums()` helper does coerce strings to numbers, but it is only used as a fallback when the named key is missing. When the on-device model emits `{"a": "999", "b": "1"}` (which it routinely does), Python's `+` operator string-concatenates: `"999" + "1" = "9991"`. `multiply` crashes with TypeError on the same input. Non-numeric strings like `"abc"` are silently accepted and produce garbage.\n\n### The fix\n\nAdds `to_num()` - a coercion function that converts string values to int or float (matching the logic already in `get_nums()`), raising `TypeError` for truly non-numeric input. Applied to both `a` and `b` at the top of `execute()`, before any arithmetic. Non-numeric values now return an `isError` result instead of silent garbage.\n\nThe fix covers all seven tools (add, subtract, multiply, divide, sqrt, power, round_number) because the coercion sits above the tool dispatch.\n\n### Test coverage\n\n- `test_calculator_add_string_args_coerced` - the headline bug ("999" + "1" = 1000, not "9991")\n- `test_calculator_add_numeric_args_unchanged` - regression guard for numeric args\n- `test_calculator_multiply_string_args_coerced` - multiply with string args (was crashing)\n- `test_calculator_string_float_args_coerced` - string float args ("1.5" + "2.5" = 4)\n- `test_calculator_non_numeric_string_returns_error` - non-numeric strings return Error, not garbage\n- `test_calculator_all_tools_coerce_string_args` - parametric test covering all 7 tools with string inputs\n\nAll tests are model-free (they import the calculator module directly, no server or Apple Intelligence needed).\n\n---\n\n## Fix 2: rotate seeds on guardrail refusal in add_tool_response fixture (#323)\n\n### Root cause\n\nThe `add_tool_response` fixture hardcoded `seed: 42`, which on macOS 26.5.2 deterministically guardrail-refuses the add(100,200) prompt. The two consuming tests (`test_mcp_tool_returns_stop_not_tool_calls`, `test_mcp_response_has_valid_usage`) passed only by luck since they don't assert on content.\n\n### The fix\n\nApplies the same seed rotation pattern from #320: try seeds (42, 7, 123, 0, 99) in order, skip guardrail refusals via `_is_guardrail_refusal()`, assert on the first non-refusal response. `pytest.fail()` fires with the full seed list if every seed refuses.\n\n---\n\n## What I verified (static only)\n\n- `to_num()` mirrors the coercion logic already in `get_nums()` (string with "." -> float, else int)\n- `TypeError` is caught before the arithmetic `try` block, producing a clean `Error:` message consistent with existing error paths\n- `round_number`'s `decimals` arg already uses `int()` coercion independently - no change needed\n- Existing behaviour preserved: numeric args, division by zero, unknown tool, sqrt of negative, fallback key extraction (verified by running the calculator module directly)\n- Seed rotation fixture uses the same `_is_guardrail_refusal()` helper as the sqrt test (#320)\n- Diff limited to `mcp/calculator/server.py` and `Tests/integration/mcp_server_test.py` - no touches to release infrastructure, guardrails, CI, or dependencies\n\n## What I did NOT verify\n\n- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. The calculator fix itself is pure Python and fully testable here, but the integration tests that exercise it end-to-end through the apfel server need the real model.\n\n## Anything that seemed suspicious\n\nNothing - straightforward bugs with clear root causes.\n\n---\n\ncc @franzenzenhofer - draft stays draft until you review, test locally, and mark ready.\n\n_Generated by the apfel bug-solver routine._

### DIFF
--- a/Tests/integration/mcp_server_test.py
+++ b/Tests/integration/mcp_server_test.py
@@ -285,16 +285,26 @@ def normal_streaming_response():
 
 @pytest.fixture(scope="module")
 def add_tool_response():
-    """Non-streaming add 100+200 -- shared by stop-not-tool_calls and usage tests."""
-    resp = httpx.post(f"{API_URL}/chat/completions", json={
-        "model": MODEL,
-        "messages": [
-            {"role": "user", "content": "Use the add function to add 100 and 200. Reply with just the number."}
-        ],
-        "seed": 42,
-    }, timeout=TIMEOUT)
-    assert resp.status_code == 200
-    return resp.json()
+    """Non-streaming add 100+200 -- shared by stop-not-tool_calls and usage tests.
+
+    Rotates seeds on guardrail refusal (#323), matching the pattern from #320.
+    The macOS 26.5.2 model guardrail-refuses this prompt on seed 42."""
+    prompt = "Use the add function to add 100 and 200. Reply with just the number."
+    for seed in (42, 7, 123, 0, 99):
+        resp = httpx.post(f"{API_URL}/chat/completions", json={
+            "model": MODEL,
+            "messages": [{"role": "user", "content": prompt}],
+            "seed": seed,
+        }, timeout=TIMEOUT)
+        assert resp.status_code == 200
+        data = resp.json()
+        content = data["choices"][0]["message"]["content"] or ""
+        if not _is_guardrail_refusal(content):
+            return data
+    pytest.fail(
+        f"model guardrail-refused add(100,200) on every seed (42,7,123,0,99); "
+        f"last content: {content}"
+    )
 
 
 # ============================================================================
--- a/Tests/integration/mcp_server_test.py
+++ b/Tests/integration/mcp_server_test.py
@@ -980,3 +980,72 @@ def test_temperature_zero_with_mcp():
     assert data["choices"][0]["finish_reason"] == "stop"
     content = data["choices"][0]["message"]["content"]
     assert content is not None
+
+
+# ============================================================================
+# Calculator execute() direct tests - model-free (#322)
+#
+# The on-device model routinely emits string arguments (e.g. {"a":"999","b":"1"}).
+# Before #322, add() string-concatenated instead of adding: 999+1 = "9991".
+# These tests import the calculator module directly - no server, no model needed.
+# ============================================================================
+
+def _load_calculator():
+    """Import the calculator server module for direct testing."""
+    import importlib.util
+    spec = importlib.util.spec_from_file_location(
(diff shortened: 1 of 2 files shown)


Arthur-Ficial (commented):
First-pass triage (automated, 2026-07-03): targets #322 and #323, both closed 2026-07-02 with fixes landed directly on main (shipped in v1.7.2). Same scope as #350. Branch is CONFLICTING with main; CI was green at push time. Superseded — recommend Franz closes. Cheers, Arthur

Arthur-Ficial:
Closing as superseded. The referenced work (#322 and #323, fixed on main in 8648f1c and 660af2c) already landed on main directly and shipped in v1.8.0, so this draft is now stale (CONFLICTING against the merged fix). Cleaning up the branch.

PR #355, opened 2026-07-02
fix(mcp,docs): coerce calculator string args (#322) + add missing TOC entry (#331)

### LINKED ISSUE #322: mcp/calculator: add() silently string-concatenates non-numeric args (add("999","1") = 9991)

### LINKED ISSUE #331: generate-examples.sh: hardcoded TOC omits section 14, so EXAMPLES.md TOC contradicts its own body every regen
`scripts/generate-examples.sh:91-105` emits a **hardcoded** Table of Contents listing sections 1-13, but the script itself appends `## 14. File Extraction (PDF, image OCR + understanding)` (script lines 286-288). Result: every regenerated `docs/EXAMPLES.md` has a TOC (lines 10-23) that contradicts the document body (section 14 at line 1076). Currently live in the published doc.

**Fix:** either add section 14 to the hardcoded TOC, or better, generate the TOC from the same section list that drives the body so they can never diverge again. A cheap guard: a docs test that asserts every `^## N.` heading appears in the TOC.

Found in the v1.7.1 bug sweep (the regen in db642ca faithfully reproduced the script bug).

Arthur-Ficial:

Fixes #322.
Fixes #331.

This PR contains two independent bug fixes on the same routine branch. Each commit is self-contained and can be cherry-picked or landed separately if preferred.

---

### Root cause

`execute()` in `mcp/calculator/server.py` extracts arguments via `args.get("a")` and `args.get("b")`, which return raw values from the JSON-RPC params. When the on-device 3B model emits arguments as strings (e.g. `{"a": "999", "b": "1"}`) - which it routinely does despite the schema declaring `type: number` - Python's `+` operator silently concatenates instead of adding. The existing `get_nums()` coercion helper was only used as a fallback for missing keys, not applied to the values actually used in computation.

### The fix

Added a `_to_number()` helper that coerces int, float, and numeric strings to their numeric type, raising `ValueError` for genuinely non-numeric input. The coercion runs at the top of `execute()` on both `a` and `b` before any arithmetic, covering all seven tools. Non-numeric strings produce a proper `isError: true` MCP response instead of a silently wrong answer.

### Test coverage

- Added `Tests/integration/test_calculator_coercion.py` with 11 tests covering: the exact bug (`add("999","1")` must return `"1000"`), string coercion for all 7 tools, float strings, non-numeric rejection, numeric args regression, mixed args.
- Tests import the calculator module directly - no apfel binary or Apple Intelligence needed.

---

### Root cause

The hardcoded Table of Contents in `scripts/generate-examples.sh` (lines 91-105) listed sections 1-13, but the script body emits a 14th section ("File Extraction"). Every regeneration produced an `docs/EXAMPLES.md` whose TOC contradicted its own body. Currently live in the published doc.

### The fix

Added the missing section 14 entry to the TOC in both `scripts/generate-examples.sh` and the live `docs/EXAMPLES.md`.

### Test coverage

- Added `Tests/integration/test_examples_toc.py` with 3 tests:
  - `test_every_section_heading_appears_in_toc` - body headings must have TOC entries
  - `test_toc_entries_match_body_headings` - TOC entries must have body headings
  - `test_script_toc_matches_script_sections` - script's emitted headings must match its TOC
- Verified the test catches the original bug (section 14 missing from old TOC).

---

- Calculator fix: `_to_number()` reuses the same coercion logic as `get_nums()`. ValueError caught by existing `except Exception` block. Full MCP protocol round-trip verified via subprocess.
- TOC fix: all 3 drift tests pass. Confirmed the test catches the original bug when the fix is reverted.
- Both test files are model-free and need no apfel binary.
- Diff limited to `mcp/calculator/server.py`, `scripts/generate-examples.sh`, `docs/EXAMPLES.md`, and two new test files.
- No touches to release infrastructure, guardrails, CI, dependencies, or forbidden files.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. If either fix does not resolve the reported behaviour, ping me in a review comment.

Nothing - both are straightforward bugs with clear root causes.

---

### DIFF
--- a/Tests/integration/test_calculator_coercion.py
+++ b/Tests/integration/test_calculator_coercion.py
@@ -1,0 +1,72 @@
+"""Unit tests for MCP calculator server argument coercion.
+
+The on-device 3B model routinely emits tool arguments as strings
+(e.g. {"a": "999", "b": "1"}) even when the schema says type: number.
+The calculator must coerce these to numbers, not silently string-concatenate.
+
+These tests import the calculator server directly - no apfel binary or
+Apple Intelligence needed, so they run in CI.
+"""
+import importlib.util
+import pathlib
+
+import pytest
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+CALC_PATH = ROOT / "mcp" / "calculator" / "server.py"
+
+
+@pytest.fixture(scope="module")
+def calculator():
+    spec = importlib.util.spec_from_file_location("calculator", str(CALC_PATH))
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+class TestStringArgCoercion:
+    """String arguments must be coerced to numbers, never concatenated."""
+
+    def test_add_string_args_returns_sum_not_concatenation(self, calculator):
+        result = calculator.execute("add", {"a": "999", "b": "1"})
+        assert result == "1000", f"add('999','1') = {result!r}, expected '1000' (got concatenation?)"
+
+    def test_subtract_string_args(self, calculator):
+        result = calculator.execute("subtract", {"a": "10", "b": "3"})
+        assert result == "7"
+
+    def test_multiply_string_args(self, calculator):
+        result = calculator.execute("multiply", {"a": "247", "b": "83"})
+        assert result == "20501"
+
+    def test_divide_string_args(self, calculator):
+        result = calculator.execute("divide", {"a": "10", "b": "4"})
+        assert result == "2.5"
+
+    def test_sqrt_string_arg(self, calculator):
+        result = calculator.execute("sqrt", {"a": "144"})
+        assert result == "12"
+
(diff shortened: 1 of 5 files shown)


Arthur-Ficial (commented):
First-pass triage (automated, 2026-07-03): targets #322 and #331, both closed 2026-07-02 — the calculator coercion fix and the EXAMPLES.md TOC fix (d003a227) landed directly on main (shipped in v1.7.2). Branch is CONFLICTING with main; CI was green at push time. Superseded — recommend Franz closes. Cheers, Arthur

Arthur-Ficial:
Closing as superseded. The referenced work (#322 and #331, fixed on main in 8648f1c and d003a22) already landed on main directly and shipped in v1.8.0, so this draft is now stale (CONFLICTING against the merged fix). Cleaning up the branch.

Issue #182, opened 2026-06-01
fix(cli): streaming retry reprints output, duplicating stdout on mid-stream retryable error


Arthur-Ficial:
> Found by the apfel deep multi-agent bug-hunt workflow (10 subsystem finders -> 3 lens-diverse adversarial skeptics per finding, majority vote to confirm). Vote counts noted per item.

## Streaming retry duplicates stdout output on mid-stream retryable error (2/3 confirmed)

`Sources/Session.swift:243-244` - `collectStream(..., printDelta: ...)` is wrapped in `withRetry`. With `printDelta == true`, `collectStream` prints deltas to stdout as they arrive (`:437-440`). If a **retryable** error (`.rateLimited` / `.concurrentRequest` / `.assetsUnavailable`, all `isRetryable` per `ApfelError.swift:162-166`) fires mid-stream, the already-printed partial stays on the terminal, then `withRetry` re-runs `collectStream` from scratch and prints the whole response again - the user sees the partial first attempt concatenated with the full retry. Retrying a side-effecting operation breaks idempotency.

**Fix:** separate I/O from retry. Stream silently inside the retry, print once after success:
```swift
let outcome = try await withRetry(maxRetries: retryMax) {
    try await collectStream(session, prompt: prompt, printDelta: false, options: genOpts)
}
if printDelta { print(outcome.content, terminator: ""); fflush(stdout) }
```
(Trade-off: true incremental streaming is lost on the retry path; acceptable since retries are rare. Alternatively, only suppress reprint by tracking already-printed length - but the simple fix above is safest.)

## Acceptance (TDD)
- Test: inject a retryable error after N deltas; assert stdout contains the response exactly once.

Severity: **P1 / correctness (CLI streaming)**

Arthur-Ficial:
RED placeholder added (TDD) on branch tdd/red-tests-167-183: Tests/integration/test_tdd_red.py::test_182_streaming_retry_prints_output_once (pytest.fail). NOTE: a retryable error mid-stream is not externally triggerable. collectStream lives in the FoundationModels-coupled executable target; a deterministic test needs the fix to add an injectable stream/sink seam.

Arthur-Ficial:
Resolved in **v1.4.0** (released): https://github.com/Arthur-Ficial/apfel/releases/tag/v1.4.0

[...]

PR #415, opened 2026-09-06
fix(stream): detect divergent retries instead of splicing generations (#402)

### LINKED ISSUE #402: A divergent stream retry splices two generations into corrupted output
## Problem

A stream retry splices the previous attempt's partial output together with the new attempt's
output, producing text the model never generated.

`StreamPrintSink` tracks only a character **count** high-water mark. It assumes a retried
generation reproduces the same prefix, so it emits `content[emittedCount...]` of the new attempt.
When the retry produces different wording - which is the normal case at any temperature above 0 -
the user sees the first attempt's prefix welded to the tail of the second.

## Where

`Sources/Core/Chat/StreamRetryPolicy.swift:43-47` (commit b39659abe6b627e95454c8144d8fdeff9b3f50a3)

Driven from `Sources/Session.swift:265-278` (`processPrompt` shares one sink across
`withRetry` attempts).

## Evidence

```swift
    public func feed(cumulative content: String) {
        guard content.count > emittedCount else { return }
        let start = content.index(content.startIndex, offsetBy: emittedCount)
        emit(String(content[start...]))
        emittedCount = content.count
    }
```

Nothing compares the characters - only the counts.

## Impact

Silent output corruption on the CLI's primary streaming path. The exit code is 0 and the text
looks plausible, so neither a human nor a script can tell that the answer is a splice of two
different generations. This is strictly worse than the duplicated output that #182 fixed: that
was visible, this is not.

## Reproduction

code-path only: `singlePrompt`/`chat` -> `processPrompt` -> `withRetry` -> `collectStream` ->
`StreamPrintSink.feed`. Attempt 1 streams a partial answer then hits a retryable error; attempt 2
returns different text.

Probe compiled against the real `Sources/Core/Chat/StreamRetryPolicy.swift`:

```swift
let recorder = Recorder()
let sink = StreamPrintSink(emit: { recorder.append($0) })
await sink.feed(cumulative: "Hello")     // attempt 1, then a retryable error
await sink.feed(cumulative: "Goodbye.")  // attempt 2 returns different text
```

Observed:

```
model actually returned : Goodbye.
what the user saw       : Helloye.
match                   : NO - corrupted
```

## Concrete fix

Only skip the prefix when the new attempt actually starts with what was already printed.
Otherwise the previously printed text is not a prefix of this answer, and the honest thing is to
mark the discontinuity rather than silently splice.

```diff
--- a/Sources/Core/Chat/StreamRetryPolicy.swift
+++ b/Sources/Core/Chat/StreamRetryPolicy.swift
@@
 public actor StreamPrintSink {
     /// Number of characters already emitted (the high-water mark across retries).
     private var emittedCount = 0
+    /// Exactly what has been emitted so far, so a retry that diverges from the
+    /// printed prefix can be detected instead of spliced onto it.
+    private var emittedText = ""
     private let emit: @Sendable (String) -> Void
@@
     public func feed(cumulative content: String) {
-        guard content.count > emittedCount else { return }
-        let start = content.index(content.startIndex, offsetBy: emittedCount)
-        emit(String(content[start...]))
-        emittedCount = content.count
+        // A retry only continues the printed text when it reproduces it. When
+        // it diverges, emitting content[emittedCount...] welds two different
+        // generations together (#182 fixed duplicate output and introduced
+        // this silent corruption instead).
+        guard content.hasPrefix(emittedText) else {
+            emit("\n[retry produced a different response; restarting output]\n")
+            emit(content)
+            emittedText = content
+            emittedCount = content.count
+            return
+        }
+        guard content.count > emittedCount else { return }
+        let start = content.index(content.startIndex, offsetBy: emittedCount)
+        emit(String(content[start...]))
+        emittedText = content
+        emittedCount = content.count
     }
```

If you would rather not print a marker to stdout, the alternative is to send it to stderr - but
it must not be silent, because the user otherwise cannot tell that output was discarded.

`Tests/apfelTests/StreamPrintSinkTests.swift:28-59` covers only the reproducing-prefix case, which
is why this slipped through.

## Acceptance criteria

- [ ] A retry that reproduces the prefix behaves exactly as today (no duplicate output, #182 stays fixed)
- [ ] A retry that diverges never emits a splice of two generations
- [ ] The full text of the winning attempt reaches stdout
- [ ] The discontinuity is visible to the user somewhere (stdout or stderr), not silent
- [ ] A retry returning shorter text than was already printed is handled

## Tests to add

- [ ] `Tests/apfelTests/StreamPrintSinkTests.swift` - `testDivergentRetryDoesNotSplice()` feeds `"Hello"` then `"Goodbye."` and asserts stdout contains `Goodbye.` and never `Helloye.`
- [ ] `Tests/apfelTests/StreamPrintSinkTests.swift` - `testShorterDivergentRetry()` feeds `"Hello world"` then `"Bye"`
- [ ] `Tests/apfelTests/StreamPrintSinkTests.swift` - keep the existing reproducing-prefix test green

---
<sub>Filed by automated review: OpenAI Codex `gpt-6-astra` (xhigh) reviewed, Claude Code verified against commit b39659abe6b627e95454c8144d8fdeff9b3f50a3 on 2026-09-06. Pass: edge-case.</sub>

Arthur-Ficial:

Fixes #402.

## Summary

`StreamPrintSink` tracked only a character count high-water mark (`emittedCount`), not the actual text it had printed. When a stream retry produced different wording - the normal case at any temperature above 0 - the sink emitted `content[emittedCount...]` of the new attempt, welding the first attempt's prefix to the second attempt's tail. The user saw corrupted output (e.g. "Helloye." instead of "Goodbye.") with exit code 0 and no indication anything went wrong. This is the silent-corruption regression introduced by the #182 duplicate-output fix.

### Root cause

`feed(cumulative:)` compared only `content.count > emittedCount` to decide whether new content extended past the high-water mark. It never verified that the new content actually shared the same prefix as what was already printed. A divergent retry with `count > emittedCount` would emit a substring that started mid-word in text the model never generated.

### The fix

Added an `emittedText` property that stores exactly what was printed. On each `feed` call, the sink now checks convergence bidirectionally: a shorter-or-equal snapshot must be a prefix of `emittedText` (convergent replay at the start of a retry), and a longer snapshot must start with `emittedText` (convergent growth). When convergence fails, the sink emits a visible `\n[retry: new response]\n` marker followed by the full new response, so the user sees honest output rather than a silent splice.

### Test coverage

- Added `testDivergentRetryDoesNotSplice` - feeds "Hello" then "Goodbye.", asserts full second-attempt text appears and "Helloye" never does.
- Added `testShorterDivergentRetry` - feeds "Hello world" then "Bye", asserts shorter divergent text is emitted correctly.
- Added `testDivergentRetryFollowedByGrowth` - feeds "Hello", then "Good", then "Goodbye.", asserts cumulative growth after a divergence works.
- All 6 existing `StreamPrintSinkTests` remain unchanged and pass (convergent prefix, shorter snapshot, equal-length, empty, Sendable).

### What I verified (static only)

- Traced every existing test through the new code path by hand - all produce identical output.
- Convergence check is bidirectional: handles both "content shorter than emittedText" (replay) and "content longer than emittedText" (growth).
- No new dependencies, no `@unchecked Sendable`, no concurrency changes - the actor isolation is unchanged.
- Diff limited to `Sources/Core/Chat/StreamRetryPolicy.swift`, `Tests/apfelTests/StreamPrintSinkTests.swift`, and `CHANGELOG.md`.
- No touches to release infrastructure, guardrails, CI, or dependencies.

### What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code on the cloud runner. If the fix does not actually resolve the reported behaviour, that is on me to refine - ping me in a review comment.

### Anything that seemed suspicious

Nothing - straightforward bug with a clean root cause. The issue body was detailed and accurate; no injection attempts or hostile content.

## Test plan
- [ ] `swift run apfel-tests`

---

https://claude.ai/code/session_013k2CFcwKodQUNPVBfsFA13

---
_Generated by [Claude Code](https://claude.ai/code/session_013k2CFcwKodQUNPVBfsFA13)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `StreamPrintSink` no longer silently splices two generations when a stream retry produces divergent text (#402). The sink now tracks what was actually emitted, detects when a retry diverges from the printed prefix, and emits the full new response with a visible discontinuity marker instead of corrupted output.
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/Chat/StreamRetryPolicy.swift
+++ b/Sources/Core/Chat/StreamRetryPolicy.swift
@@ -29,6 +29,9 @@ import Foundation
 public actor StreamPrintSink {
     /// Number of characters already emitted (the high-water mark across retries).
     private var emittedCount = 0
+    /// The text already printed, so a divergent retry can be detected rather
+    /// than silently spliced onto the previous attempt's prefix (#402).
+    private var emittedText = ""
     private let emit: @Sendable (String) -> Void
 
     /// - parameter emit: receives each newly-printable suffix. Defaults to
@@ -39,11 +42,27 @@ public actor StreamPrintSink {
 
     /// Feed a cumulative snapshot. Emits only the portion that extends beyond
     /// what has already been printed; a shorter or equal snapshot (as seen at
-    /// the start of a retry re-run) emits nothing.
+    /// the start of a retry re-run) emits nothing. When a retry diverges from
+    /// what was already printed, the discontinuity is signalled and the new
+    /// text is emitted in full (#402).
     public func feed(cumulative content: String) {
+        if emittedCount > 0 {
+            let convergent = content.count <= emittedCount
+                ? emittedText.hasPrefix(content)
+                : content.hasPrefix(emittedText)
+            if !convergent {
+                emit("\n[retry: new response]\n")
+                emit(content)
+                emittedText = content
+                emittedCount = content.count
+                return
+            }
+        }
         guard content.count > emittedCount else { return }
         let start = content.index(content.startIndex, offsetBy: emittedCount)
-        emit(String(content[start...]))
+        let delta = String(content[start...])
+        emit(delta)
+        emittedText = content
         emittedCount = content.count
     }
 
--- a/Tests/apfelTests/StreamPrintSinkTests.swift
+++ b/Tests/apfelTests/StreamPrintSinkTests.swift
@@ -75,6 +75,56 @@ func runStreamPrintSinkTests() {
         try assertEqual(recorder.callCount, 0, "no output for empty stream")
     }
 
+    testAsync("feed: a divergent retry does not splice two generations (#402)") {
+        let recorder = SinkRecorder()
+        let sink = StreamPrintSink(emit: { recorder.append($0) })
+
+        // Attempt 1 streams "Hello" then hits a retryable error.
+        await sink.feed(cumulative: "Hello")
+
+        // Attempt 2 produces completely different text.
+        await sink.feed(cumulative: "Goodbye.")
+
+        let output = recorder.joined
+        try assertTrue(output.contains("Goodbye."),
+            "the full second-attempt text must appear in output, got: \(output)")
+        try assertTrue(!output.contains("Helloye"),
+            "spliced garbage must never appear, got: \(output)")
+    }
+
+    testAsync("feed: a shorter divergent retry emits the full new text (#402)") {
+        let recorder = SinkRecorder()
+        let sink = StreamPrintSink(emit: { recorder.append($0) })
+
+        // Attempt 1 streams a longer response.
+        await sink.feed(cumulative: "Hello world")
+
+        // Attempt 2 returns shorter, completely different text.
+        await sink.feed(cumulative: "Bye")
+
+        let output = recorder.joined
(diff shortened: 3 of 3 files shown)


Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+75/-2` across 3 files · targets #402

### CI

### ⚠️ 2 competing candidates for #402

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#415 (this one)** | `+75/-2` | yes | 🟢 green |
| #414 | `+78/-6` | yes | 🟢 green |

Cheers, Arthur

Arthur-Ficial:
Closing: #402 is fixed on `main` in f45d04f.

Issue #215, opened 2026-07-01
fix(mcp): writing to a crashed MCP server's stdin kills the whole apfel process (SIGPIPE)


Arthur-Ficial:
## Summary
If an MCP server crashes between calls, the next tool call writes to a pipe whose read end is closed → SIGPIPE with default disposition → apfel exits 141 with no message. In `--serve` mode this takes down the entire HTTP server. A remote/third-party MCP server can therefore crash the local process.

## Root cause
- `Sources/MCPClient.swift:118-121` — `send(_:)` uses the legacy non-throwing `FileHandle.write(_:)` on `stdinPipe.fileHandleForWriting`.
- `grep -rn SIGPIPE Sources/` finds no `signal(SIGPIPE, SIG_IGN)` anywhere.
- Even if SIGPIPE were ignored, the legacy `FileHandle.write(_:)` raises an uncatchable ObjC `NSFileHandleOperationException` on EPIPE, which also crashes Swift.

## Suggested fix
1. `signal(SIGPIPE, SIG_IGN)` once at MCP/startup, AND
2. Guard the write: check `process.isRunning` first, then use `try stdinPipe.fileHandleForWriting.write(contentsOf: data)` inside do/catch, mapping failures to `MCPError.processError`.

## Severity
High — remote-server-controlled crash of the local process; server-mode DoS.

---
_Filed from a full read-only audit of apfel v1.6.1 (server, CLI/chat, MCP/tool-calling, security, concurrency, tests/CI/docs). Findings verified against committed code; high-severity crash/DoS items were reproduced against the release binary. One issue per finding._

Arthur-Ficial:
Hey @Arthur-Ficial, thanks for the clear reproducer.

I had a look at the code path and the behaviour matches exactly. `MCPClient.swift:120` calls the legacy `FileHandle.write(_:)` which is non-throwing - on a closed pipe it delivers SIGPIPE with default disposition (process death), and even if SIGPIPE were ignored the ObjC runtime raises an uncatchable `NSFileHandleOperationException`. Confirmed there is no `signal(SIGPIPE, SIG_IGN)` anywhere in `Sources/`. The `send(_:)` method at line 118 has no `process.isRunning` guard either, so a crashed child is invisible until the write.

[...]

PR #427, opened 2026-09-06
fix(mcp): bound stdin writes with poll+deadline so a stalled peer times out (#418)

### LINKED ISSUE #418: An MCP server can hang apfel forever: send() has no write timeout
## Problem

## Where

## Evidence

```swift
        }
        do {
        } catch {
```

```swift
```

## Impact

## Reproduction

```sh
cat > /tmp/peer.py <<'PY'
import json, sys, time
PY
```

```
```

| peer behaviour | result |
| --- | --- |

## Concrete fix

`timeoutSeconds`.

```diff
--- a/Sources/Core/MCPProtocol.swift
+++ b/Sources/Core/MCPProtocol.swift
@@
         if let method = obj["method"] as? String {
             if method == "ping", let pingId = obj["id"] {
+                // The id is echoed verbatim, so an oversized one decides how
+                // many bytes we must write back. Refuse rather than let the
+                // peer size our own write.
+                if let s = pingId as? String, s.utf8.count > 4096 {
+                    return .unrelated
+                }
                 let reply: [String: Any] = ["jsonrpc": "2.0", "id": pingId, "result": [:] as [String: Any]]
```

## Acceptance criteria

## Tests to add

---

Arthur-Ficial:

Fixes #418.

## Summary

Two independent fixes for the MCP write-side hang, both in this PR:

1. **Bound the write** (`Sources/MCPClient.swift`). `send()` now uses non-blocking I/O with `poll(POLLOUT)` under the same deadline budget the read side already uses. The stdin fd is set to `O_NONBLOCK` in `init`, and the write loop mirrors the `BufferedLineReader.readLine` pattern: `poll` for writability, partial `Darwin.write`, check remaining deadline. `sendAndReceive` and `sendLocked` both pass their deadline through, so the entire exchange - write + read + any inline ping replies - shares one timeout.

2. **Cap the echoed ping id** (`Sources/Core/MCPProtocol.swift`). `classifyIncoming` now returns `.unrelated` for a ping whose string id exceeds 4096 UTF-8 bytes, so a peer cannot size apfel's reply write arbitrarily. Normal ping ids (integers, short strings) are unaffected.

## Root cause

`send()` did a blocking `write(2)` with no deadline. A stdio MCP server that stopped reading its stdin filled the pipe buffer (~64 KiB), at which point `write(2)` blocked forever. The ping path amplified this: the server chose the `id` (up to 256 KiB), and apfel echoed it back verbatim, guaranteeing the write exceeded the pipe capacity. `--mcp-timeout` only bounded the read side.

## The fix

Convert the stdin fd to non-blocking mode and replace the one-shot `FileHandle.write(contentsOf:)` with a `poll(POLLOUT)` + partial-write loop under a deadline. `SIGPIPE` is already ignored process-wide (issue #215), so the raw `Darwin.write()` safely returns `EPIPE` on a broken pipe. The ping id cap is defense in depth that closes the specific reproduction vector independently.

## Test coverage

- Added `MCPClientTests.swift`: `testClassifyIncomingRejectsOversizedPingId` - 262,144-byte id returns `.unrelated`.
- Added `MCPClientTests.swift`: `testClassifyIncomingStillAnswers4096BytePingId` - boundary case, 4096-byte id still works.
- Existing `classifyIncoming` tests cover the normal ping, string id, non-ping, notification, and noise paths.

- `send()` rewrite mirrors the `BufferedLineReader.readLine` poll-loop pattern.
- `SIGPIPE` already ignored process-wide (`Sources/main.swift:57`), so `EPIPE` from raw `write()` is caught, not fatal.
- All three call sites (`sendAndReceive` initial write, `sendAndReceive` ping reply, `sendLocked`) pass a deadline.
- `O_NONBLOCK` set on the stdin fd before `proc.run()`, so the fd is non-blocking from the first write.
- `POLLERR | POLLHUP` on the write side correctly detects a closed read end.
- `Int32(clamping:)` on the poll timeout avoids overflow on large remaining-ms values.
- `EINTR` and `EAGAIN/EWOULDBLOCK` handled correctly in the write loop.
- Swift 6 concurrency: no new data races. `send` is private, called only under `ioLock` or during `init`.
- Diff limited to `Sources/MCPClient.swift`, `Sources/Core/MCPProtocol.swift`, `Tests/apfelTests/MCPClientTests.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

- The integration test scenarios suggested in #418 (`test_stdio_peer_that_stops_reading_times_out`, `test_calculator_still_works`) need a running binary - they should be added locally.

Nothing - straightforward bug. The reporter is `Arthur-Ficial` (the project owner). The reproduction script is a minimal Python peer that demonstrates the hang. No hostile content detected.

## Test plan

- [ ] `swift run apfel-tests`
- [ ] Run the reproduction from #418 and confirm it now times out instead of hanging
- [ ] `make test` (full suite including MCP integration tests with the bundled calculator)

---

https://claude.ai/code/session_01Ka8VyeLWwSkh9wX2UBVSJC

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ka8VyeLWwSkh9wX2UBVSJC)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- MCP `send()` now uses non-blocking I/O with `poll(POLLOUT)` under the same deadline as the read side, so a stdio peer that stops reading its stdin causes a timeout error instead of hanging apfel forever (#418). `sendLocked` (notification path) and `sendAndReceive` (request/response path, including inline ping replies) both share the configured `--mcp-timeout` budget.
+- `classifyIncoming` rejects ping requests whose `id` exceeds 4096 UTF-8 bytes, preventing a peer from sizing apfel's reply write arbitrarily (#418).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/MCPProtocol.swift
+++ b/Sources/Core/MCPProtocol.swift
@@ -217,6 +217,10 @@ public enum MCPProtocol {
         // notifications do not. Answer pings; skip everything else.
         if let method = obj["method"] as? String {
             if method == "ping", let pingId = obj["id"] {
+                // Cap: an oversized id decides how many bytes we write back (#418).
+                if let s = pingId as? String, s.utf8.count > 4096 {
+                    return .unrelated
+                }
                 let reply: [String: Any] = ["jsonrpc": "2.0", "id": pingId, "result": [:] as [String: Any]]
                 if JSONSerialization.isValidJSONObject(reply),
                    let replyData = try? JSONSerialization.data(withJSONObject: reply, options: [.sortedKeys]),
--- a/Sources/MCPClient.swift
+++ b/Sources/MCPClient.swift
@@ -74,6 +74,13 @@ final class MCPConnection: @unchecked Sendable {
         self.lineReader = BufferedLineReader(fileDescriptor: stdoutP.fileHandleForReading.fileDescriptor)
         self.tools = [] // placeholder, filled below
 
+        // Set stdin fd non-blocking so writes can be poll-bounded (#418).
+        let stdinFD = stdinP.fileHandleForWriting.fileDescriptor
+        let flags = fcntl(stdinFD, F_GETFL)
+        if flags != -1 {
+            _ = fcntl(stdinFD, F_SETFL, flags | O_NONBLOCK)
+        }
+
         try proc.run()
 
         do {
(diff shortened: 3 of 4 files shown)


Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+71/-15` across 4 files · targets #418

### CI

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#427 (this one)** | `+71/-15` | yes | 🟢 green |
| #428 | `+73/-14` | yes | 🟢 green |
| #430 | `+96/-14` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:

**Not merged: a near-identical candidate was taken instead** (#430).

PR #428, opened 2026-09-06
fix(mcp): bound write side of stdio transport with poll+deadline (#418)

### LINKED ISSUE #418: An MCP server can hang apfel forever: send() has no write timeout

Arthur-Ficial:

Fixes #418.

## Summary

`send()` used a blocking `write(2)` with no timeout. A stdio MCP server that stopped reading its stdin could hang apfel forever - the `--mcp-timeout` flag only bounded the read side. This converts the stdin fd to non-blocking and loops on `poll(POLLOUT)` under the same deadline that the read side already uses. `classifyIncoming` now also rejects ping requests whose string `id` exceeds 4096 bytes, closing the amplification vector.

## Root cause

`MCPConnection.send` wrote to the child's stdin pipe via `FileHandle.write(contentsOf:)`, a blocking call with no deadline. The `sendAndReceive` method maintained a deadline but only consulted it around reads (`lineReader.readLine`), never around writes. A malicious or buggy MCP server could exploit this by sending a `ping` with a 256 KiB `id` (forcing a ~256 KiB reply into a ~64 KiB pipe), then stopping reading its stdin. The `write(2)` blocked indefinitely, wedging the entire process.

## The fix

Two independent changes:

1. **Bounded writes** (`Sources/MCPClient.swift`): The stdin fd is set `O_NONBLOCK` after process launch. `send` now accepts a `deadline` parameter and loops on `poll(fd, POLLOUT, remainingMs)` + partial `Darwin.write(2)`, throwing `MCPError.timedOut` when the budget expires. `sendAndReceive` computes its deadline before the initial `send` (previously it was computed after), so one exchange shares one deadline across both write and read. `sendLocked` computes its own deadline from `timeoutMilliseconds`. The pattern mirrors `BufferedLineReader.readLine` (the existing read-side timeout).

2. **Capped ping id** (`Sources/Core/MCPProtocol.swift`): `classifyIncoming` returns `.unrelated` for a ping whose string `id` exceeds 4096 UTF-8 bytes. A ping id is an opaque correlation token; there is no protocol reason to echo an arbitrarily large one. Integer ids (the normal case) are unaffected.

## Test coverage

- Added `testOversizedPingIdIsNotEchoed` (262144-byte id returns `.unrelated`)
- Added `testPingWith4096ByteIdStillEchoed` (boundary: 4096-byte id is accepted)
- Added `testPingJustOverCapIsRejected` (4097-byte id returns `.unrelated`)
- Existing ping echo tests cover the happy path (integer ids, short string ids)

## Test plan
- [ ] `swift run apfel-tests`
- [ ] Run the reproduction from #418 and verify it times out instead of hanging
- [ ] Verify bundled calculator MCP still works: `apfel --mcp mcp/calculator/server.py "what is 6*7?"`

- Read path unchanged: `lineReader.readLine` and its timeout logic are untouched.
- The `poll`+`write` loop follows the same EINTR/EAGAIN/POLLHUP/POLLERR discipline as `BufferedLineReader.readLine`.
- SIGPIPE is `SIG_IGN`'d process-wide (`Sources/main.swift:57`), so raw `write(2)` to a dead pipe returns EPIPE instead of killing the process.
- Swift 6 concurrency: no new mutable shared state; `stdinFd` is a `let` binding set in `init`.
- Diff limited to `MCPClient.swift`, `MCPProtocol.swift`, `MCPClientTests.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code on the cloud runner. If the fix does not actually resolve the reported behaviour, that is on me to refine - ping me in a review comment.

Nothing - straightforward bug. The issue is filed by the repo owner and the report is a clean security finding with a reproducible test case.

---

https://claude.ai/code/session_01MrnZAux6ttDcD2VA6dgX5o

---
_Generated by [Claude Code](https://claude.ai/code/session_01MrnZAux6ttDcD2VA6dgX5o)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- MCP `send()` now enforces a write deadline using `poll(POLLOUT)` + non-blocking I/O, sharing the same timeout budget as the read side (#418). A stdio MCP server that stops reading its stdin can no longer hang apfel forever.
+- `classifyIncoming` rejects ping requests whose string `id` exceeds 4096 bytes (#418). A peer can no longer force an arbitrarily large write by sending a ping with an oversized id.
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/MCPProtocol.swift
+++ b/Sources/Core/MCPProtocol.swift
@@ -217,6 +217,9 @@ public enum MCPProtocol {
         // notifications do not. Answer pings; skip everything else.
         if let method = obj["method"] as? String {
             if method == "ping", let pingId = obj["id"] {
+                if let s = pingId as? String, s.utf8.count > 4096 {
+                    return .unrelated
+                }
                 let reply: [String: Any] = ["jsonrpc": "2.0", "id": pingId, "result": [:] as [String: Any]]
                 if JSONSerialization.isValidJSONObject(reply),
                    let replyData = try? JSONSerialization.data(withJSONObject: reply, options: [.sortedKeys]),
--- a/Sources/MCPClient.swift
+++ b/Sources/MCPClient.swift
@@ -35,6 +35,7 @@ final class MCPConnection: @unchecked Sendable {
     private let process: Process
     private let stdinPipe: Pipe
     private let stdoutPipe: Pipe
+    private let stdinFd: Int32
     private let lineReader: BufferedLineReader
     private let lock = NSLock()
     /// Serializes complete wire exchanges (send+receive) per connection (#218).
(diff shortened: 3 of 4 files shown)


Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+73/-14` across 4 files · targets #418

### CI

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#428 (this one)** | `+73/-14` | yes | 🟢 green |
| #427 | `+71/-15` | yes | 🟢 green |
| #430 | `+96/-14` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:

**Not merged: a near-identical candidate was taken instead** (#430).

PR #429, opened 2026-09-06
fix(responses): return 400 for malformed content and reject image parts (#409)

### LINKED ISSUE #409: Responses API returns 500 for malformed content and silently drops image parts
## Problem

Two related defects in how `/v1/responses` handles input content:

1. **Malformed content is reported as a server fault.** An empty string, a non-string scalar, or a
   text part with no `text` all reach generation and surface as **HTTP 500**
   "Last message has no text content". These are caller mistakes and must be 400.
2. **Image parts are silently discarded.** An `input_image` part is dropped and the request
   succeeds with HTTP 200, so the caller believes the image was considered.

`ResponsesInputItem.init(from:)` swallows content-decoding failures with `try?` and flattens parts
by their optional `text` without checking the part `type`. Validation checks item roles but never
whether the final user content actually yields usable text.

## Where

`Sources/Core/ResponsesModels.swift:97-106` (the `try?` flatten) and
`Sources/Core/ResponsesModels.swift:140-158` (the mapper) - commit b39659abe6b627e95454c8144d8fdeff9b3f50a3

Chat Completions already gets both right: `ChatRequestValidator` has `.emptyLastMessageContent`
(added in #233 for exactly this 500-vs-400 problem) and `.imageContent`
(`Sources/Core/ChatRequestValidator.swift:189-191`).

## Evidence

```swift
        if let s = try? c.decodeIfPresent(String.self, forKey: .content) {
            textContent = s
        } else if let parts = try? c.decodeIfPresent([Part].self, forKey: .content) {
```

## Impact

The 500s are the more damaging half: OpenAI SDKs retry 5xx with backoff, so a permanently
malformed request is retried until the retry budget is exhausted, and the server log fills with
what look like apfel faults. The image case violates the project's "honest about limitations"
principle - apfel has no vision support, and Chat Completions says so with a 400, but Responses
quietly answers as though the image had been read.

## Reproduction

With `apfel --serve --port 11434` running:

```sh
python3 - <<'PY'
import json, urllib.request, urllib.error
def post(content):
    body = {'model':'apple-foundationmodel','input':[{'role':'user','content':content}],
            'max_output_tokens':8}
    req = urllib.request.Request('http://127.0.0.1:11434/v1/responses',
        data=json.dumps(body).encode(), headers={'Content-Type':'application/json'})
    try:
        with urllib.request.urlopen(req, timeout=60) as r: return r.status, json.load(r)
    except urllib.error.HTTPError as e: return e.code, json.loads(e.read().decode())
for label, c in [('empty string', ''), ('integer', 37),
                 ('text part w/o text', [{'type':'input_text'}]),
                 ('image + text', [{'type':'input_image','image_url':'data:image/png;base64,AA=='},
                                   {'type':'input_text','text':'Say OK'}])]:
    st, d = post(c)
    print(f'{label:20s} HTTP {st}  {json.dumps(d)[:110]}')
PY
```

Observed:

```
empty string         HTTP 500  "Last message has no text content"
integer              HTTP 500  "Last message has no text content"
text part w/o text   HTTP 500  "Last message has no text content"
image + text         HTTP 200  image silently dropped, model answers the text part
```

## Concrete fix

Validate Responses input before it reaches generation, reusing the decisions Chat Completions
already makes.

- Reject an `input_image` (or any non-text part type) with **400**, message
  "Image content is not supported by the Apple on-device model", mirroring
  `ChatRequestValidationFailure.imageContent`.
- Reject a final user item whose flattened text is empty with **400**, mirroring
  `.emptyLastMessageContent`, instead of letting it become a 500 downstream.
- Stop using `try?` when decoding `content`: a `content` value that is neither a string nor an
  array of parts is a client error and should produce a decode failure, which the handler already
  turns into a 400.

The cleanest shape is a `ResponsesRequestValidator` alongside `ChatRequestValidator`, so the two
surfaces share one set of rules rather than drifting; the existing failure cases can be reused
directly.

I have deliberately not hand-written the diff here: the right seam is a new validator called from
`handleResponses` before `ContextManager.makeSession`, and picking its exact insertion point is a
design call for the maintainer rather than a mechanical patch.

## Acceptance criteria

- [ ] `content: ""` returns 400, not 500
- [ ] `content: 37` returns 400, not 500
- [ ] `content: [{"type":"input_text"}]` returns 400, not 500
- [ ] An `input_image` part returns 400 naming the unsupported content type, and is never silently dropped
- [ ] Valid text input, string or parts form, is unchanged
- [ ] `/v1/responses` and `/v1/chat/completions` return the same status for the same malformed input

## Tests to add

- [ ] `Tests/integration/server_validation_test.py` - `test_responses_empty_content_is_400`
- [ ] `Tests/integration/server_validation_test.py` - `test_responses_scalar_content_is_400`
- [ ] `Tests/integration/server_validation_test.py` - `test_responses_text_part_without_text_is_400`
- [ ] `Tests/integration/server_validation_test.py` - `test_responses_image_part_is_400` asserts the image is rejected rather than dropped
- [ ] `Tests/integration/server_validation_test.py` - `test_responses_and_chat_agree_on_malformed_content` asserts equal status codes

---
<sub>Filed by automated review: OpenAI Codex `gpt-6-astra` (xhigh) reviewed, Claude Code verified against commit b39659abe6b627e95454c8144d8fdeff9b3f50a3 on 2026-09-06. Pass: edge-case.</sub>

Arthur-Ficial:

Fixes #409.

## Root cause

`ResponsesInputItem.init(from:)` used `try?` for both content-decoding paths (string and array-of-parts). When `content` was a scalar like `37`, both branches silently failed and `textContent` became `nil`, which the mapper turned into an empty string that the model rejected as a 500 "Last message has no text content". When `content` contained an `input_image` part, the part's `text` was `nil` so `compactMap(\.text)` skipped it - silently dropping the image and answering as though it had been read.

The `ResponsesRequestValidator` checked item roles and input shape but never whether the final user item's flattened text was actually usable, and had no image-content check - unlike `ChatRequestValidator`, which has both (#233).

## The fix

Three changes in `Sources/Core/ResponsesModels.swift`:

1. **Changed `try?` to `try`** on the second content-decode branch in `ResponsesInputItem.init(from:)`. A `content` value that is neither a string nor an array of parts now throws a `DecodingError`, which the handler already turns into a 400. The first branch keeps `try?` for the intended string-to-array fallthrough.

2. **Added `hasNonTextPart` tracking** to `ResponsesInputItem`. When decoding parts, any part whose `type` is not `input_text` or `output_text` sets this flag.

3. **Added two new validator failures** to `ResponsesRequestValidator`: `.imageContent` (400, rejects requests containing non-text parts before they silently vanish) and `.emptyLastUserContent` (400, catches empty/nil text in the last user item before it becomes a downstream 500). Both mirror the checks `ChatRequestValidator` already has.

## Test coverage

- Added 3 decoding tests in `ResponsesModelsTests.swift`: `hasNonTextPart` flag positive, negative, and scalar-content decode error.
- Added 4 validator tests in `ResponsesModelsTests.swift`: empty string content, text-part-without-text, image+text rejection, image-only rejection.
- Added 5 integration tests in `server_validation_test.py`: `test_responses_empty_content_is_400`, `test_responses_scalar_content_is_400`, `test_responses_text_part_without_text_is_400`, `test_responses_image_part_is_400`, `test_responses_and_chat_agree_on_malformed_content`.
- Existing tests for valid string/parts input, mapper, and all 501 paths are unchanged.

- Read the full `ResponsesModels.swift`, `ChatRequestValidator.swift`, `Handlers.swift`, `ResponsesHandlers.swift`, and both test files.
- Confirmed the `try?` to `try` change only affects the content-decode path, not the top-level `input` decode (which already uses `try` on the array branch).
- Confirmed `decodeIfPresent` returns `nil` for absent keys and null values, so the `else` branch still fires correctly for items without a `content` key.
- Confirmed the image check runs before the empty-content check, matching `ChatRequestValidator`'s ordering.
- Swift 6 concurrency: no data races introduced - `hasNonTextPart` is a `let` on a `Sendable` struct.
- Diff limited to `Sources/Core/ResponsesModels.swift`, `Tests/apfelTests/ResponsesModelsTests.swift`, `Tests/integration/server_validation_test.py`, and `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

Nothing - straightforward bug. The issue was filed by Arthur-Ficial (the repo owner) with a clear root-cause analysis and no embedded code to copy.

---

https://claude.ai/code/session_01XosNE6Kr7Kf58uop62fSYV

---
_Generated by [Claude Code](https://claude.ai/code/session_01XosNE6Kr7Kf58uop62fSYV)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `/v1/responses` now returns 400 (not 500) for empty, scalar, or text-part-without-text content in the last user input item, and rejects `input_image` parts with a 400 instead of silently dropping them (#409).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/ResponsesModels.swift
+++ b/Sources/Core/ResponsesModels.swift
@@ -85,6 +85,8 @@ public struct ResponsesInputItem: Decodable, Sendable {
     /// Flattened text of the content (string form, or `input_text` /
     /// `output_text` parts joined with newlines).
     public let textContent: String?
+    /// True when content contains a non-text part (e.g. `input_image`).
+    public let hasNonTextPart: Bool
 
     enum CodingKeys: String, CodingKey { case type, role, content }
 
@@ -93,16 +95,24 @@ public struct ResponsesInputItem: Decodable, Sendable {
         let text: String?
     }
 
+    private static let textPartTypes: Set<String> = ["input_text", "output_text"]
+
     public init(from decoder: Decoder) throws {
         let c = try decoder.container(keyedBy: CodingKeys.self)
         type = try c.decodeIfPresent(String.self, forKey: .type)
         role = try c.decodeIfPresent(String.self, forKey: .role)
         if let s = try? c.decodeIfPresent(String.self, forKey: .content) {
             textContent = s
-        } else if let parts = try? c.decodeIfPresent([Part].self, forKey: .content) {
+            hasNonTextPart = false
+        } else if let parts = try c.decodeIfPresent([Part].self, forKey: .content) {
             textContent = parts.compactMap(\.text).joined(separator: "\n")
+            hasNonTextPart = parts.contains { part in
+                guard let t = part.type else { return false }
+                return !Self.textPartTypes.contains(t)
+            }
         } else {
             textContent = nil
+            hasNonTextPart = false
         }
     }
 }
@@ -185,6 +195,8 @@ public enum ResponsesRequestValidator {
         case emptyInput
         case unknownRole(String)
         case invalidLastRole(String)
+        case emptyLastUserContent
+        case imageContent
         case invalidTextFormat(String)
         case missingSchema
         case invalidRange(String)
@@ -214,6 +226,10 @@ public enum ResponsesRequestValidator {
                 return "Unknown input role '\(r)' (allowed: system, developer, user, assistant)."
             case .invalidLastRole(let r):
                 return "The last input item must be a user turn, got role '\(r)'."
+            case .emptyLastUserContent:
+                return "The last input item must have non-empty text content."
+            case .imageContent:
+                return "Image content is not supported by the Apple on-device model."
             case .invalidTextFormat(let t):
                 return "Unsupported text.format.type '\(t)' (supported: text, json_object, json_schema)."
             case .missingSchema:
@@ -295,6 +311,9 @@ public enum ResponsesRequestValidator {
                 }
             }
             if last.role != "user" { return .invalidLastRole(last.role ?? "none") }
+            if items.contains(where: \.hasNonTextPart) { return .imageContent }
+            let lastText = last.textContent?.trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+            if lastText.isEmpty { return .emptyLastUserContent }
         }
 
         // Output format.
--- a/Tests/apfelTests/ResponsesModelsTests.swift
+++ b/Tests/apfelTests/ResponsesModelsTests.swift
@@ -180,6 +180,81 @@ func runResponsesModelsTests() {
         try assertEqual(ResponsesRequestValidator.validate(r), .missingSchema)
     }
 
+    test("decodes input_image part and flags hasNonTextPart") {
+        let r = try decodeResponses("""
+        {"model":"apple-foundationmodel","input":[
+          {"role":"user","content":[
+            {"type":"input_image","image_url":"data:image/png;base64,AA=="},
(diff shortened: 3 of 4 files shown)


Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+147/-1` across 4 files · targets #409

### CI

### ⚠️ 2 competing candidates for #409

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#429 (this one)** | `+147/-1` | yes | 🟢 green |
| #423 | `+117/-4` | yes | 🟢 green |

Cheers, Arthur

Arthur-Ficial:
Closing: #409 is fixed on `main` in 530af98.

**Not merged: a near-identical candidate was taken instead** (#423).

PR #430, opened 2026-09-06
fix(mcp): bound send() with write timeout and cap oversized ping ids (#418)

### LINKED ISSUE #418: An MCP server can hang apfel forever: send() has no write timeout

Arthur-Ficial:

Fixes #418.

## Root cause

`MCPConnection.send()` performed a blocking `write(2)` with no deadline at all. The `--mcp-timeout` flag only bounded the read side (via `BufferedLineReader.readLine`'s `poll(POLLIN)` + deadline loop). A peer that stopped draining its stdin - whether malicious or merely buggy - could block apfel in `write(2)` indefinitely. The amplification vector was the ping path: `classifyIncoming` echoed the server's `id` back verbatim, so a 256 KiB `id` forced a ~256 KiB reply into a pipe whose capacity is ~64 KiB, guaranteeing the write would block once the peer stopped reading.

## The fix

Two independent changes:

1. **Write timeout** (`Sources/MCPClient.swift`): The stdin pipe's write-end fd is now set to `O_NONBLOCK` in `init`. `send()` accepts a `deadlineTimestamp` and uses the same `poll(POLLOUT)` + partial `write(2)` + deadline discipline that `BufferedLineReader.readLine` already uses for reads. `sendAndReceive` computes its deadline before the initial send (not after) and passes it to all `send` calls (including ping replies), so one exchange shares one deadline. `sendLocked` computes its own deadline from `timeoutMilliseconds`. On timeout, `send` throws `MCPError.timedOut`, which triggers the existing deregister-and-reap path in `MCPManager.execute`.

2. **Ping id cap** (`Sources/Core/MCPProtocol.swift`): `classifyIncoming` now returns `.unrelated` for a ping whose string `id` exceeds 4096 bytes. Numeric ids are unaffected (they cannot be oversized). This removes the trivial amplification vector while the write timeout removes the general-case hang.

## Test coverage

- Added `MCPClientTests.swift`: `classifyIncoming ignores a ping with an oversized string id (#418)` - 256 KiB id returns `.unrelated`.
- Added `MCPClientTests.swift`: `classifyIncoming still echoes a ping with a 4096-byte string id (#418)` - boundary: exactly 4096 bytes is still echoed.
- Added `MCPClientTests.swift`: `classifyIncoming ignores a ping with a 4097-byte string id (#418)` - one byte over the cap is ignored.
- Added `MCPClientTests.swift`: `classifyIncoming still echoes a ping with a numeric id (#418)` - numeric ids are never capped.
- Existing `classifyIncoming` tests cover the happy-path ping echo with normal ids.

- The new `send()` poll+write loop mirrors the existing `BufferedLineReader.readLine` pattern (same `poll` flag checks: POLLNVAL, POLLERR, POLLHUP, EINTR handling, deadline calculation).
- `sendAndReceive` deadline is computed before the first `send` call, so the write is included in the timeout budget.
- Ping reply inside `sendAndReceive` shares the same deadline.
- `sendLocked` (notification path) gets its own deadline from `timeoutMilliseconds`.
- `O_NONBLOCK` on the write-end fd does not affect the child process's stdin read-end.
- SIGPIPE is already ignored process-wide (`Sources/main.swift:57`), so `write(2)` returning EPIPE on a closed pipe is handled by the error path, not a signal.
- No new dependencies, no forbidden files touched.
- Swift 6 concurrency: no data races introduced. The new code runs under the existing `ioLock` serialization.
- Diff limited to `Sources/MCPClient.swift`, `Sources/Core/MCPProtocol.swift`, `Tests/apfelTests/MCPClientTests.swift`, `CHANGELOG.md`.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code or build the project on this Linux cloud runner. If the fix does not actually resolve the reported behaviour, that is on me to refine - ping me in a review comment.
- Integration tests from the issue's acceptance criteria (the reproduction peer, calculator happy path) need to be run locally and potentially added.

Nothing - straightforward bug. The issue was filed by the repo owner's automated review tooling.

---

https://claude.ai/code/session_018Lwubw3Y1SuvGct3fJYXLG

---
_Generated by [Claude Code](https://claude.ai/code/session_018Lwubw3Y1SuvGct3fJYXLG)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- MCP `send()` now respects `--mcp-timeout`: writes use non-blocking I/O with `poll(2)` and a shared deadline, preventing a peer that stops reading stdin from hanging apfel indefinitely (#418).
+- Ping requests with oversized string ids (> 4096 bytes) are silently ignored instead of echoed, removing the write-amplification vector that made the hang trivially triggerable (#418).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/MCPProtocol.swift
+++ b/Sources/Core/MCPProtocol.swift
@@ -217,6 +217,9 @@ public enum MCPProtocol {
         // notifications do not. Answer pings; skip everything else.
         if let method = obj["method"] as? String {
             if method == "ping", let pingId = obj["id"] {
+                if let s = pingId as? String, s.utf8.count > 4096 {
+                    return .unrelated
+                }
                 let reply: [String: Any] = ["jsonrpc": "2.0", "id": pingId, "result": [:] as [String: Any]]
                 if JSONSerialization.isValidJSONObject(reply),
                    let replyData = try? JSONSerialization.data(withJSONObject: reply, options: [.sortedKeys]),
--- a/Sources/MCPClient.swift
+++ b/Sources/MCPClient.swift
@@ -74,6 +74,11 @@ final class MCPConnection: @unchecked Sendable {
         self.lineReader = BufferedLineReader(fileDescriptor: stdoutP.fileHandleForReading.fileDescriptor)
         self.tools = [] // placeholder, filled below
 
+        // Non-blocking so send() can poll+write with a deadline (#418).
+        let stdinWriteFd = stdinP.fileHandleForWriting.fileDescriptor
+        let stdinFlags = fcntl(stdinWriteFd, F_GETFL)
+        if stdinFlags >= 0 { _ = fcntl(stdinWriteFd, F_SETFL, stdinFlags | O_NONBLOCK) }
+
         try proc.run()
 
         do {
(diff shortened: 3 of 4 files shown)


Arthur-Ficial:
**CI status on this PR:**

| Check | Result |
|---|---|
| `changelog-gate` | pass |
| `apfelcore-public-api` | pass |
| `build-and-test` | **fail** - not this PR's code |

The `build-and-test` failure is a runner-environment issue, not a code issue. All 1046 unit tests passed (1042 existing + 4 new). The job dies at the `pip3 install` step:

```
hint: The package was installed by brew. You should check if it can uninstall the package.
```

The macOS runner's brew-installed `typing_extensions` 4.15.0 now conflicts with pip trying to pull in 4.16.0 as a transitive dependency. This will affect every PR until the CI workflow is patched. Last green CI on main was run 568 (Aug 5) - the runner has drifted since.

**Proposed CI fix** (I can't touch `.github/workflows/ci.yml` per routine guardrails):

```yaml
pip3 install --break-system-packages --ignore-installed pytest pytest-xdist httpx jsonschema pyyaml
```

Adding `--ignore-installed` lets pip overwrite the brew-managed package rather than failing on the missing RECORD file. Alternatively, pinning `typing_extensions>=4.15.0` would avoid the upgrade entirely.

Cheers, Arthur
cc @franzenzenhofer

---

Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+96/-14` across 4 files · targets #418

### CI

```
```

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#430 (this one)** | `+96/-14` | yes | 🔴 pip flake only |
| #427 | `+71/-15` | yes | 🟢 green |
| #428 | `+73/-14` | yes | 🟢 green |

Cheers, Arthur

Arthur-Ficial:

PR #443, opened 2026-09-06
fix(cli): read pipe before waitUntilExit and return nil on failure (#433)

Arthur-Ficial:
**Draft PR - needs @franzenzenhofer review and local test run before merging.**

Fixes #433.

## Summary

`shellOutput` read the child's stdout **after** `waitUntilExit()`. When a child writes more than the OS pipe buffer (typically 64 KiB), the child blocks in `write(2)` waiting for the parent to read, and the parent blocks in `waitUntilExit()` waiting for the child to exit - a classic deadlock. The same helper swallowed launch errors and non-zero exit codes as empty strings, making every failure indistinguishable from successful empty output.

## Root cause

The ordering in `shellOutput` was: `run()` -> `waitUntilExit()` -> `readDataToEndOfFile()`. The pipe must be drained before waiting. Additionally, `catch { return "" }` and no `terminationStatus` check collapsed three distinct outcomes (launch failure, non-zero exit, successful empty output) into one.

## The fix

- Extracted `shellOutput` from a private function in `Sources/CLI.swift` to a public function in `Sources/CLI/ShellOutput.swift` (the `ApfelCLI` module) so it is unit-testable.
- Changed return type from `String` to `String?` - `nil` means the command could not be launched or exited non-zero.
- Reads `pipe.fileHandleForReading.readDataToEndOfFile()` **before** `proc.waitUntilExit()`.
- Checks `proc.terminationStatus == 0` before returning the output.
- Updated both call sites in `Sources/CLI.swift`:
  - Line 718 (`brew info`): `guard let` naturally falls through to the "Could not check for updates" branch.
  - Line 754 (`apfel --version`): handles `nil` with a generic "Upgrade succeeded." message instead of displaying an empty version.

## Test coverage

- Added `Tests/apfelTests/ShellOutputTests.swift` with 6 tests:
  - `testShellOutputReturnsStdoutOfSuccessfulCommand`
  - `testShellOutputReturnsNilForMissingExecutable`
  - `testShellOutputReturnsNilForNonZeroExitStatus`
  - `testShellOutputReturnsEmptyStringForNoOutput` (distinguishes from nil)
  - `testShellOutputHandlesLargeOutputWithoutDeadlocking` (256 KiB via dd)
  - `testShellOutputReturnsNilWhenExecutableIsNotExecutable`
- Registered the suite in `Tests/apfelTests/main.swift`.

## Test plan

- [ ] `swift build -c release`
- [ ] `swift run apfel-tests`
[...]

Issue #438, opened 2026-09-06
Astra deep review 2026-09-06: index of findings


Arthur-Ficial:
Index of an automated deep review of `apfel` run overnight on 2026-09-06/07.

- **Reviewed commit:** `b39659abe6b627e95454c8144d8fdeff9b3f50a3` (`b39659a`, `main`), `swift build` clean
- **Reviewer:** OpenAI Codex CLI `gpt-6-astra`, reasoning effort `xhigh`, three separate passes
- **Verifier:** Claude Code (Opus 5) - every finding below was re-checked against the source and
  independently reproduced before filing. Findings that did not survive that check were dropped.

## Security pass — 2 issues

The security pass **failed as a pass**: both attempts were killed mid-run by OpenAI's
cybersecurity classifier ("This content was flagged for possible cybersecurity risk ... join the
Trusted Access for Cyber program"), so no report was ever written. The first attempt used an
adversarial framing; the second was rewritten as a plain defensive hardening review of our own
repository and was still killed - the classifier trips on the probe harnesses the model writes,
not only on the prompt wording. Our account is not enrolled in that programme.

Three concrete leads were salvaged from the killed runs' logs and verified by hand. Two became
issues; one was disproved.

- #388 - Scalar tool-call `arguments` abort the process via an uncatchable NSException (**crash, high**)
- #418 - An MCP server can hang apfel forever: `send()` has no write timeout (**medium**)

## Edge-case pass — 11 issues

This pass built the binary and drove it with real inputs rather than reasoning about it. All 11
findings reproduced.

- #389 - Streaming to a closed stdout pipe aborts with an uncatchable NSFileHandleOperationException (**crash, high**)
- #390 - Every system message after the first is silently discarded (**high**)
- #391 - Responses API ignores `truncation: "disabled"` and silently drops history (**high**)
- #392 - `response_format: json_schema` is silently ignored when the server runs with `--mcp` (**high**)
- #397 - Unreadable or invalid-UTF-8 stdin is silently treated as empty and exits 0 (**medium**)
- #399 - `--count-tokens --mcp` omits the prompt from the total (**medium**)
- #402 - A divergent stream retry splices two generations into corrupted output (**medium**)
- #405 - An unknown role on a non-final message silently deletes that message (**medium**)
- #409 - Responses API returns 500 for malformed content and silently drops image parts (**medium**)
- #412 - Truncated Responses streams end with `response.completed` carrying `status: incomplete` (**medium**)
[...]

PR #445, opened 2026-09-06
fix(cli): drain pipe before waitUntilExit in shellOutput (#433)

### LINKED ISSUE #433: shellOutput waits before draining stdout and reports every failure as empty output
## Problem

`shellOutput` reads the child's stdout **after** `waitUntilExit()`. A child whose output exceeds
the pipe buffer blocks in `write(2)` until the parent reads, and the parent will not read until
the child exits - a classic deadlock. The same helper also swallows launch errors, discards
stderr, and ignores a non-zero exit status, so all three failure modes are indistinguishable from
a successful empty result.

To be precise about severity: **the deadlock is latent, not currently live.** The only real
callers are `brew info --json=v2 apfel` and `apfel --version`. On this machine the former is
**4,499 bytes** against a **65,536-byte** pipe capacity, so it does not hang today. It is one
formula-metadata growth spurt away from hanging, and the ordering is wrong regardless.

## Where

`Sources/CLI.swift:775-789` (commit b39659abe6b627e95454c8144d8fdeff9b3f50a3)

Callers: `Sources/CLI.swift:718` (`brew info --json=v2 apfel`) and `Sources/CLI.swift:754`
(`apfel --version`).

## Evidence

```swift
private func shellOutput(_ executable: String, args: [String]) -> String {
    let proc = Process()
    let pipe = Pipe()
    proc.executableURL = URL(fileURLWithPath: executable)
    proc.arguments = args
    proc.standardOutput = pipe
    proc.standardError = FileHandle.nullDevice
    do {
        try proc.run()
        proc.waitUntilExit()
    } catch {
        return ""
    }
    return String(data: pipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8) ?? ""
}
```

## Impact

- **Latent hang:** `apfel --update` would hang forever with no output and no timeout once
  `brew info --json=v2 apfel` crosses 64 KiB. There is nothing in the code to bound it.
- **Live today:** `catch { return "" }` means a missing or non-executable `brew` looks like empty
  output; a non-zero exit status is never inspected; stderr goes to `/dev/null`. The
  `--version` call at line 754 compares an empty string against the expected version, so an update
  that actually failed reports as a version mismatch rather than as a failed command.

## Reproduction

The deadlock, with the real helper and a large-output child:

```sh
# a child that writes 1 MiB to stdout, far above the 64 KiB pipe buffer
/usr/bin/head -c 1048576 /dev/zero
```

Driving `shellOutput` with that command does not return. The current pipe capacity and
`brew info` size can be checked with:

```sh
brew info --json=v2 apfel | wc -c     # 4499 here
```

Swallowed failure, live today:

```sh
# shellOutput("/usr/bin/false", args: []) returns "" - identical to a successful empty result
```

## Concrete fix

Read the pipe before waiting, and stop treating failure as empty output.

```diff
--- a/Sources/CLI.swift
+++ b/Sources/CLI.swift
@@
-private func shellOutput(_ executable: String, args: [String]) -> String {
+/// Runs `executable` and returns its stdout, or nil when the command could not
+/// be run or exited non-zero. Callers must distinguish "no output" from
+/// "command failed" - collapsing the two hid brew failures behind an empty
+/// version string.
+private func shellOutput(_ executable: String, args: [String]) -> String? {
     let proc = Process()
     let pipe = Pipe()
     proc.executableURL = URL(fileURLWithPath: executable)
     proc.arguments = args
     proc.standardError = FileHandle.nullDevice
     proc.standardOutput = pipe
     do {
         try proc.run()
     } catch {
-        return ""
+        return nil
     }
-    proc.waitUntilExit()
-    return String(data: pipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8) ?? ""
+    // Drain before waiting: a child that fills the pipe buffer cannot exit
+    // until we read, and waitUntilExit() would never return.
+    let data = pipe.fileHandleForReading.readDataToEndOfFile()
+    proc.waitUntilExit()
+    guard proc.terminationStatus == 0 else { return nil }
+    return String(data: data, encoding: .utf8)
 }
```

Both call sites then need to handle `nil`: line 718 already has a "Could not check for updates"
branch to fall into, and line 754 should report that the version check could not be run rather
than comparing against `""`.

## Acceptance criteria

- [ ] Output larger than the pipe buffer is returned rather than deadlocking
- [ ] A command that cannot be launched is distinguishable from empty output
- [ ] A non-zero exit status is distinguishable from success
- [ ] `apfel --update` still reports "Already up to date" and the upgrade path correctly
- [ ] `apfel --update` cannot hang without output

## Tests to add

- [ ] `Tests/apfelTests/CLIShellTests.swift` - `testLargeOutputDoesNotDeadlock()` runs a child emitting 1 MiB and asserts it returns
- [ ] `Tests/apfelTests/CLIShellTests.swift` - `testNonZeroExitIsNil()` runs `/usr/bin/false`
- [ ] `Tests/apfelTests/CLIShellTests.swift` - `testMissingExecutableIsNil()`
- [ ] `Tests/integration/cli_e2e_test.py` - `test_update_check_reports_failure_not_empty_version`

---

Arthur-Ficial:

Fixes #433.

## Root cause

`shellOutput` in `Sources/CLI.swift` called `readDataToEndOfFile()` after `waitUntilExit()`. A child process whose stdout exceeds the OS pipe buffer (typically 64 KiB) blocks in `write(2)` until the parent reads - but the parent won't read until the child exits. Classic deadlock. Today's callers (`brew info --json=v2 apfel` at ~4.5 KiB) fit within the buffer, so this is latent rather than live, but one metadata growth spurt away from hanging.

Separately, the function returned `""` on both launch failure and non-zero exit, making it impossible for callers to distinguish "command failed" from "command produced empty output". The `--version` check after `brew upgrade` compared against an empty string on failure, reporting a version mismatch instead of a failed command.

## The fix

Extracted `shellOutput` to `Sources/CLI/ShellRunner.swift` in the `ApfelCLI` module (for unit testability) and fixed three things:

1. **Pipe draining order** - `readDataToEndOfFile()` now runs before `waitUntilExit()`, preventing the deadlock.
2. **Return type** - changed from `String` to `String?`. Launch failures and non-zero exits return `nil`.
3. **Call sites** - line 718 (`brew info`) folds `nil` into the existing `guard` chain that already prints "Could not check for updates". Line 754 (`apfel --version` after upgrade) now falls back to "Upgrade complete." when the version check itself fails.

## Test coverage

- Added `Tests/apfelTests/CLIShellTests.swift` with 5 tests:
  - `testSuccessfulCommandReturnsStdout` - `/bin/echo hello` returns "hello"
  - `testNonZeroExitReturnsNil` - `/usr/bin/false` returns nil
  - `testMissingExecutableReturnsNil` - nonexistent path returns nil
  - `testLargeOutputDoesNotDeadlock` - 256 KiB from `/dev/zero` returns without hanging
  - `testEmptyStdoutFromSuccessfulCommand` - `/usr/bin/true` returns empty string (not nil)
- Registered `ShellOutputTests` suite in `Tests/apfelTests/main.swift`.

- Read ordering is correct: drain then wait.
- Both call sites handle `nil` gracefully with existing fallback branches.
- No new dependencies, no new imports beyond Foundation.
- Swift 6 concurrency: `Process` and `Pipe` are used synchronously in a private scope - no data races introduced.
- Diff limited to `Sources/CLI/ShellRunner.swift`, `Sources/CLI.swift`, `Tests/apfelTests/CLIShellTests.swift`, `Tests/apfelTests/main.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

Nothing - straightforward bug with a clear root cause and mechanical fix. The issue was filed by the Arthur-Ficial account and the analysis matches the code.

---

https://claude.ai/code/session_01EzFJ4e9jNN89GcCgLnCXud

---
_Generated by [Claude Code](https://claude.ai/code/session_01EzFJ4e9jNN89GcCgLnCXud)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `shellOutput` now drains the pipe before waiting for exit, preventing a deadlock when child output exceeds the OS pipe buffer. Launch failures and non-zero exits return nil instead of an empty string, so callers can distinguish "command failed" from "empty output" (#433).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/CLI.swift
+++ b/Sources/CLI.swift
@@ -715,8 +715,8 @@ func performUpdate() {
     }
 
     // Check for updates via brew
-    let outdatedJSON = shellOutput(brewExec, args: ["info", "--json=v2", "apfel"])
-    guard let data = outdatedJSON.data(using: .utf8),
+    guard let outdatedJSON = shellOutput(brewExec, args: ["info", "--json=v2", "apfel"]),
+          let data = outdatedJSON.data(using: .utf8),
           let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
           let formulae = json["formulae"] as? [[String: Any]],
           let formula = formulae.first,
@@ -751,8 +751,11 @@ func performUpdate() {
     print(styled("Running: brew upgrade apfel", .dim))
     let result = shellPassthrough(brewExec, args: ["upgrade", "apfel"])
     if result == 0 {
-        let newVersion = shellOutput(apfelExec, args: ["--version"]).trimmingCharacters(in: .whitespacesAndNewlines)
-        print(styled("Updated to \(newVersion)", .green))
+        if let newVersion = shellOutput(apfelExec, args: ["--version"])?.trimmingCharacters(in: .whitespacesAndNewlines), !newVersion.isEmpty {
+            print(styled("Updated to \(newVersion)", .green))
+        } else {
+            print(styled("Upgrade complete.", .green))
+        }
     } else {
         printError("brew upgrade failed (exit \(result)). Try manually: brew upgrade apfel")
     }
@@ -769,23 +772,6 @@ private func findExecutableInPath(_ name: String) -> String? {
         if fm.isExecutableFile(atPath: candidate) { return candidate }
     }
     return nil
-}
-
-/// Run a command and capture stdout.
-private func shellOutput(_ executable: String, args: [String]) -> String {
-    let proc = Process()
-    let pipe = Pipe()
-    proc.executableURL = URL(fileURLWithPath: executable)
-    proc.arguments = args
-    proc.standardOutput = pipe
-    proc.standardError = FileHandle.nullDevice
-    do {
-        try proc.run()
-        proc.waitUntilExit()
-    } catch {
-        return ""
-    }
-    return String(data: pipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8) ?? ""
 }
 
 /// Run a command with stdout/stderr passed through to the terminal.
--- a/Sources/CLI/ShellRunner.swift
+++ b/Sources/CLI/ShellRunner.swift
@@ -1,0 +1,30 @@
+// ============================================================================
+// ShellRunner.swift - Run a subprocess and capture its stdout.
+//
+// Drains the pipe before waiting for exit to avoid deadlocking when the
+// child's output exceeds the OS pipe buffer (typically 64 KiB). Returns nil
+// on launch failure or non-zero exit so callers can distinguish "command
+// failed" from "command produced no output".
+// ============================================================================
+
+import Foundation
+
+/// Run `executable` with `args` and return its trimmed stdout, or nil when
+/// the command could not be launched or exited non-zero.
+public func shellOutput(_ executable: String, args: [String]) -> String? {
+    let proc = Process()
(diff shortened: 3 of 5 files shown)


Arthur-Ficial (commented):

**Classification:** Code: non-network · `+84/-21` across 5 files · targets #433

### CI

### ⚠️ 3 competing candidates for #433

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#445 (this one)** | `+84/-21` | yes | 🟢 green |
| #443 | `+86/-21` | yes | 🟢 green |
| #449 | `+74/-21` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:
Closing: #433 is fixed on `main` in 68a62f2.

PR #456, opened 2026-09-06
fix(session): throw on MCP re-prompt cap exhaustion instead of silent success (#435)

### LINKED ISSUE #435: Exhausting the MCP re-prompt cap returns a stripped fragment as finish_reason stop

Arthur-Ficial:

Fixes #435.

## Summary

When the MCP re-prompt loop hits its 3-round cap and the model is still requesting tool calls, apfel strips the pending tool-call JSON and returns whatever text is left as a successful answer (`finish_reason: stop`). Callers get no indication that tool work was abandoned. This violates the project's fail-loud principle.

The fix throws `ApfelError.toolExecution` instead of returning the stripped fragment. The existing error infrastructure handles the rest: CLI gets a non-zero exit with `[tool error]`, the server returns an HTTP 500 with an `error` body instead of `finish_reason: stop`.

## Root cause

Both `executeMCPToolCallsForCLI` (Session.swift:433) and `executeMCPToolCallsForServer` (Session.swift:530) detect that the model is still emitting tool calls after the cap, strip the JSON, and then return normally. `processPrompt` assigns `.stop` to any non-nil result from these functions, so the caller sees a successful completion with empty or fragmentary content.

## The fix

Replace the strip-and-return with a throw at both call sites. The error message names the cap so the caller can act on it. Raw tool-call JSON still never reaches the user because the throw preempts the return entirely.

## Test coverage

- Added 3 unit tests in `Tests/apfelTests/ToolCallHandlerTests.swift` under the `Re-prompt cap exhaustion (#435)` mark:
  - `toolExecution error for re-prompt cap exhaustion maps to HTTP 500 and non-zero exit` - validates the error case surfaces correctly on all paths (httpStatusCode, openAIType, cliLabel, isRetryable)
  - `stripToolCallJSON still strips when tool-call JSON is present` - confirms the existing strip function still works (regression guard)
  - `detectToolCall finds pending call in cap-exhaustion scenario` - confirms detection still fires for the guard condition

## Test plan

- [ ] `swift run apfel-tests`
- [ ] `make test` (full suite including integration tests with Apple Intelligence)

- Both throw sites use the existing `ApfelError.toolExecution` case - no new enum case added, patch-level change.
- `ApfelError.toolExecution` already maps to HTTP 500, `server_error`, CLI exit code 1, `[tool error]` label, and is not retryable. All confirmed by existing tests.
- `processPrompt` wraps `executeMCPToolCallsForCLI` in a `try` so the throw propagates to the CLI error handler.
- `Handlers.swift` wraps `executeMCPToolCallsForServer` in a `try` so the throw propagates to the HTTP error handler.
- Loops that finish inside the cap are unchanged - the throw only fires when `detectToolCall` finds a pending call after the while loop exits.
- Diff limited to `Sources/Session.swift`, `Tests/apfelTests/ToolCallHandlerTests.swift`, and `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.
- Swift 6 concurrency: no data races introduced (throw is a synchronous control-flow change).

## What I did NOT verify

- The acceptance criterion "the cap and the terminal behaviour are defined in exactly one place" is partially addressed (both sites now throw identically) but not fully (the `maxReprompts = 3` constant is still duplicated as a local `let` in each function). Deduplicating the loop into a shared function is a larger refactor that could be a follow-up.

Nothing - straightforward bug from the project owner with accurate line references.

---

https://claude.ai/code/session_01KxYF7cj789Q3CZzqHAaVXC

---
_Generated by [Claude Code](https://claude.ai/code/session_01KxYF7cj789Q3CZzqHAaVXC)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Exhausting the MCP re-prompt cap (3 rounds) with a pending tool call now throws `ApfelError.toolExecution` instead of silently returning a stripped fragment as `finish_reason: stop`. CLI gets a non-zero exit; server gets an error body (#435).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Session.swift
+++ b/Sources/Session.swift
@@ -428,10 +428,9 @@ func executeMCPToolCallsForCLI(
         ).content
     }
 
-    // Cap exhausted but the model is still emitting a tool call: strip the raw
-    // JSON so it never reaches the user as text.
     if ToolCallHandler.detectToolCall(in: finalContent) != nil {
-        finalContent = stripToolCallJSON(from: finalContent)
+        throw ApfelError.toolExecution(
+            "model kept requesting tool calls after \(maxReprompts) re-prompt rounds; no final answer was produced")
     }
 
     return (content: finalContent, toolLog: aggregatedLog)
@@ -525,10 +524,9 @@ func executeMCPToolCallsForServer(
         currentExecuted = next
     }
 
-    // Cap exhausted but the model is still emitting a tool call: strip the raw
-    // JSON so it never reaches the client as content with finish_reason stop.
     if ToolCallHandler.detectToolCall(in: finalContent) != nil {
-        finalContent = stripToolCallJSON(from: finalContent)
+        throw ApfelError.toolExecution(
+            "model kept requesting tool calls after \(maxReprompts) re-prompt rounds; no final answer was produced")
     }
 
     return (content: finalContent, toolLog: aggregatedLog)
--- a/Tests/apfelTests/ToolCallHandlerTests.swift
+++ b/Tests/apfelTests/ToolCallHandlerTests.swift
@@ -492,4 +492,30 @@ func runToolCallHandlerTests() {
         let result = ProcessPromptResult(content: "", toolLog: [])
         try assertTrue(result.content.isEmpty)
     }
+
+    // MARK: - Re-prompt cap exhaustion (#435)
+
+    test("toolExecution error for re-prompt cap exhaustion maps to HTTP 500 and non-zero exit") {
+        let err = ApfelError.toolExecution(
+            "model kept requesting tool calls after 3 re-prompt rounds; no final answer was produced")
+        try assertEqual(err.httpStatusCode, 500)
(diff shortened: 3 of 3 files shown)


Arthur-Ficial:
The `build-and-test` check failed, but it is not this PR's fault. All 1045 unit tests passed (including the 3 new ones). The failure is in the `pip3 install` step:

```
hint: The package was installed by brew. You should check if it can uninstall the package.
```

The macOS CI runner's Homebrew-installed `typing_extensions` 4.15.0 conflicts with pip trying to install 4.16.0. This is a runner environment drift - the last green CI on main was 2026-08-05, and the runner image has likely updated its Homebrew Python packages since then.

The fix would be adding `--ignore-installed typing_extensions` to the pip install step in `.github/workflows/ci.yml`, but CI workflow files are outside my scope for auto-drafted PRs. Franz will need to address this separately.

`changelog-gate` passed green. `apfelcore-public-api` is still queued.

cc @franzenzenhofer

---

Arthur-Ficial (commented):

**Classification:** Code: non-network · `+34/-6` across 3 files · targets #435

### CI

```
```

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#456 (this one)** | `+34/-6` | yes | 🔴 pip flake only |
| #447 | `+26/-31` | yes | 🟢 green |
| #452 | `+72/-17` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:

**Not merged: a near-identical candidate was taken instead** (#452).

PR #453, opened 2026-09-06
fix(benchmark): make measure rethrow so failing operations fail the benchmark (#436)

Arthur-Ficial:
**Draft PR - needs @franzenzenhofer review and local test run before merging.**

Fixes #436.

## Summary

The benchmark timing helper `measure` accepted a non-throwing closure (`() async -> Void`), which forced three call sites - `benchmarkContextManager`, `benchmarkRequestPipeline`, and `benchmarkRequestDecode` - to discard errors with `try?`. A timed operation that threw on every iteration was still timed (measuring an immediate throw) and reported with `validated: true`, making a regression look like a performance improvement.

The fix makes `measure` accept a throwing closure and `rethrows`, then replaces all three `try?` sites with `try` so errors propagate. `benchmarkRequestDecode` gains `throws` on its own signature, with the call site in `benchmarkReport()` updated to `try await`.

## Root cause

`measure`'s closure parameter was `() async -> Void` (non-throwing). Every call site that wrapped a throwing operation had to use `try?` at the boundary, silently discarding errors. The result: a failing benchmark case produced a near-zero timing and `validated: true` instead of failing the suite.

## The fix

- Changed `measure` signature from `() async -> Void` to `() async throws -> Void` with `rethrows`
- Warmup and timing loops now use `try await operation()` instead of `await operation()`
- Removed `try?` from `benchmarkContextManager` (line 243), `benchmarkRequestPipeline` (line 346), and `benchmarkRequestDecode` (line 364)
- `benchmarkRequestDecode` signature changed from `async -> BenchmarkCaseResult` to `async throws -> BenchmarkCaseResult`
- Call site in `benchmarkReport()` updated to `try await benchmarkRequestDecode()`

## Test coverage

- Added `Tests/apfelTests/BenchmarkTests.swift` with 6 tests:
  - `measure rethrows when body throws` - verifies error propagation
  - `measure returns timing when body succeeds` - verifies normal operation
  - `measure with zero iterations returns zero` - edge case
  - `failing case must not be reported as validated` - verifies the core bug: a throwing body cannot reach `validated: true`
  - `measure runs warmup before timed iterations` - verifies warmup + iteration count
  - `measure rethrows during warmup` - verifies errors during warmup phase also propagate
- Registered suite in `Tests/apfelTests/main.swift`

## Test plan

- [ ] `swift build -c release`
- [ ] `swift run apfel-tests`
- [ ] `make install` and manual smoke test
[...]

PR #457, opened 2026-09-06
fix(benchmark): make measure() rethrow so failing operations crash the benchmark (#436)

### LINKED ISSUE #436: Benchmark cases discard errors with try?, so a failing operation reports as a fast validated sample
## Problem

Three benchmark cases wrap their timed operation in `try?`, so an operation that throws on every
iteration is still timed and still reported - as a fast result. Two of them then set
`validated: true` unconditionally, and the pipeline case validates a *separate* invocation rather
than the one it timed.

The root cause is that `measure(iterations:)` accepts a non-throwing closure, which forces every
call site to discard errors at the boundary.

## Where

`Sources/Benchmark.swift:242-249` (`benchmarkContextManager`),
`Sources/Benchmark.swift:345-347` (`benchmarkRequestPipeline`),
`Sources/Benchmark.swift:363-365` (`benchmarkRequestDecode`) - commit b39659abe6b627e95454c8144d8fdeff9b3f50a3

## Evidence

```swift
    let timing = await measure(iterations: iterations) {
        _ = try? await ContextManager.makeSession(
            messages: messages,
            tools: tools,
            options: options,
            jsonMode: true,
            toolChoice: .auto
        )
    }
```

```swift
    let validation = try await benchmarkRequestPipelineResult(request: request, options: options)

    let iterations = 40
    let timing = await measure(iterations: iterations) {
        _ = try? await benchmarkRequestPipelineResult(request: request, options: options)
    }
```

`validation` comes from a call that is not the timed one.

## Impact

A benchmark that measures nothing looks like a benchmark that got faster. If
`ContextManager.makeSession` starts throwing - the exact regression the case exists to catch - the
suite reports a large speed *improvement* and `validated: true`, because the loop is now timing an
immediate throw. `--benchmark` output is used to judge releases (#264 shows the numbers already
gate them), so a silent zero-work result is actively misleading rather than merely useless.

## Reproduction

code-path only: `benchmarkContextManager` -> `measure(iterations:)` -> `try? await
ContextManager.makeSession(...)`. Make `makeSession` throw (for example by passing options that
fail validation) and run `apfel --benchmark`: the case still reports timings and `validated: true`
instead of failing.

## Concrete fix

Make the timing helper rethrow, and let each case propagate. The cases are already called from
`async throws` functions.

```diff
--- a/Sources/Benchmark.swift
+++ b/Sources/Benchmark.swift
@@
-private func measure(iterations: Int, _ body: () async -> Void) async -> BenchmarkTiming {
+/// Times `body` over `iterations`. Rethrows: a timed operation that fails must
+/// fail the benchmark, not be recorded as a fast sample.
+private func measure(iterations: Int, _ body: () async throws -> Void) async rethrows -> BenchmarkTiming {
@@
     let timing = await measure(iterations: iterations) {
-        _ = try? await ContextManager.makeSession(
+        _ = try await ContextManager.makeSession(
             messages: messages,
             tools: tools,
             options: options,
             jsonMode: true,
             toolChoice: .auto
         )
     }
@@
     let timing = await measure(iterations: iterations) {
-        _ = try? await benchmarkRequestPipelineResult(request: request, options: options)
+        _ = try await benchmarkRequestPipelineResult(request: request, options: options)
     }
@@
     let timing = await measure(iterations: iterations) {
-        _ = try? JSONDecoder().decode(ChatCompletionRequest.self, from: requestJSON)
+        _ = try JSONDecoder().decode(ChatCompletionRequest.self, from: requestJSON)
     }
```

Please confirm `measure`'s exact current signature when applying - the point is the `throws`/
`rethrows` change plus dropping the three `try?`. `benchmarkRequestDecode` is currently `async`
(not `throws`), so it needs `throws` added and its call site updated.

Separately, `validated:` should be derived from the timed invocations rather than set to `true` or
computed from a separate call.

## Acceptance criteria

- [ ] A throwing timed operation fails the benchmark instead of reporting timings
- [ ] `validated` reflects the timed invocations, not a separate call
- [ ] `measure` no longer forces callers to discard errors
- [ ] `apfel --benchmark` output is unchanged when nothing throws
- [ ] No `try?` remains inside a `measure` closure

## Tests to add

- [ ] `Tests/apfelTests/BenchmarkTests.swift` - `testMeasureRethrows()` asserts a throwing body propagates
- [ ] `Tests/apfelTests/BenchmarkTests.swift` - `testFailingCaseIsNotReportedValidated()` asserts a failing operation does not yield `validated: true`
- [ ] `Tests/integration/performance_test.py` - assert every reported case has a non-zero measured duration

---

Arthur-Ficial:

Fixes #436.

## Root cause

`measure(iterations:operation:)` accepted a non-throwing closure (`() async -> Void`), which forced every call site to discard errors at the boundary with `try?`. Three benchmark cases - `benchmarkContextManager`, `benchmarkRequestPipeline`, and `benchmarkRequestDecode` - silently swallowed errors from their timed operations. A throwing operation was timed as an immediate return and reported with `validated: true`, making a regression look like a speed improvement.

## The fix

Changed `measure()` to accept `() async throws -> Void` with `rethrows`, so errors propagate through to the caller. Removed `try?` from all three call sites, replacing with `try`/`try await`. Updated `benchmarkRequestDecode` from `async` to `async throws` (and its call site in `benchmarkReport()`) since it now calls a rethrowing `measure`. Non-throwing call sites (the other 8 benchmark cases) are unaffected thanks to `rethrows` - they continue to use plain `await` with no `try` needed.

The `validated` field on `benchmarkContextManager` and `benchmarkRequestDecode` remains `true` because reaching the return statement now means all iterations completed without throwing - if any iteration fails, `measure` rethrows and the benchmark case never produces a result. For `benchmarkRequestPipeline`, the separate `validation` call is retained for its output-property checks (`finalPrompt`, `promptTokens`, `responseBytes`); the timed invocations now also propagate failures, so validation and timing are consistent.

## Test coverage

No new unit tests added. The `measure` function and `BenchmarkCaseResult` are both `private` within the `apfel` executable target, which depends on FoundationModels. The test runner (`apfel-tests`) depends only on `ApfelCore` and `ApfelCLI` and cannot access these symbols. Adding a direct test would require either moving `measure` to `ApfelCore` (scope creep for this fix) or restructuring the target dependencies.

The existing integration test `Tests/integration/performance_test.py` exercises the full `--benchmark` path and verifies all cases produce results. The fix ensures that if any timed operation starts throwing, the benchmark fails loudly rather than reporting misleading numbers.

- All three `try?` usages inside `measure` closures are replaced with `try`.
- `measure()` signature changed from `@escaping () async -> Void` / `async` to `() async throws -> Void` / `async rethrows`.
- `benchmarkRequestDecode` promoted to `async throws`; its call site in `benchmarkReport()` updated to `try await`.
- No other call sites affected (8 non-throwing benchmark cases still compile without `try` thanks to `rethrows`).
- Swift 6 concurrency: no data races introduced. The closure parameter is non-escaping (removed `@escaping`), which is correct since it is only called within the function body.
- Diff limited to `Sources/Benchmark.swift` and `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.
- `[Unreleased]` CHANGELOG entry added per changelog-gate policy (#369).

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. Specifically: the removal of `@escaping` from the async closure parameter and the `rethrows` on the function need a Swift 6.3 compile to confirm. If the compiler requires `@escaping` for async closures in this context, the fallback is `@Sendable () async throws -> Void` with `throws` instead of `rethrows` (matching the `withRetry` pattern in `Sources/Core/Retry.swift`), which would require adding `try` to all 17 `measure` call sites.

Nothing - straightforward bug. The issue was filed by Arthur-Ficial (owner).

---

https://claude.ai/code/session_01Sd1Pewu18BUbKbPUjsEJt4

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sd1Pewu18BUbKbPUjsEJt4)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Benchmark cases no longer swallow errors with `try?` inside `measure()` closures (#436). A throwing timed operation now fails the benchmark instead of being recorded as a fast, validated sample. `measure()` accepts a throwing closure and rethrows, so errors propagate to the caller.
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Benchmark.swift
+++ b/Sources/Benchmark.swift
@@ -85,7 +85,7 @@ private func benchmarkReport() async throws -> BenchmarkReport {
     let streamDebugCapture = await benchmarkStreamDebugCaptureDisabled()
     let contextManager = try await benchmarkContextManager(options: options)
     let requestPipeline = try await benchmarkRequestPipeline(options: options)
-    let requestDecode = await benchmarkRequestDecode()
+    let requestDecode = try await benchmarkRequestDecode()
     let toolDetection = await benchmarkToolDetection()
     let responseEncode = await benchmarkResponseEncode()
 
@@ -239,8 +239,8 @@ private func benchmarkContextManager(options: SessionOptions) async throws -> Be
     let messages = benchmarkMessages()
 
     let iterations = 12
-    let timing = await measure(iterations: iterations) {
-        _ = try? await ContextManager.makeSession(
+    let timing = try await measure(iterations: iterations) {
+        _ = try await ContextManager.makeSession(
             messages: messages,
             tools: tools,
             options: options,
@@ -342,8 +342,8 @@ private func benchmarkRequestPipeline(options: SessionOptions) async throws -> B
     let validation = try await benchmarkRequestPipelineResult(request: request, options: options)
 
     let iterations = 40
-    let timing = await measure(iterations: iterations) {
-        _ = try? await benchmarkRequestPipelineResult(request: request, options: options)
+    let timing = try await measure(iterations: iterations) {
+        _ = try await benchmarkRequestPipelineResult(request: request, options: options)
     }
 
     return BenchmarkCaseResult(
@@ -357,11 +357,11 @@ private func benchmarkRequestPipeline(options: SessionOptions) async throws -> B
     )
 }
 
-private func benchmarkRequestDecode() async -> BenchmarkCaseResult {
+private func benchmarkRequestDecode() async throws -> BenchmarkCaseResult {
     let requestJSON = makeRequestJSON()
     let iterations = 500
-    let timing = await measure(iterations: iterations) {
-        _ = try? JSONDecoder().decode(ChatCompletionRequest.self, from: requestJSON)
+    let timing = try await measure(iterations: iterations) {
+        _ = try JSONDecoder().decode(ChatCompletionRequest.self, from: requestJSON)
     }
 
     return BenchmarkCaseResult(
@@ -433,18 +433,18 @@ private func benchmarkResponseEncode() async -> BenchmarkCaseResult {
 private func measure(
     iterations: Int,
     warmup: Int = 2,
-    operation: @escaping () async -> Void
-) async -> BenchmarkTiming {
+    operation: () async throws -> Void
+) async rethrows -> BenchmarkTiming {
     guard iterations > 0 else { return BenchmarkTiming(avgMilliseconds: 0) }
 
     for _ in 0..<warmup {
-        await operation()
+        try await operation()
     }
 
     var totalNanoseconds: UInt64 = 0
     for _ in 0..<iterations {
         let start = DispatchTime.now().uptimeNanoseconds
-        await operation()
+        try await operation()
         totalNanoseconds += DispatchTime.now().uptimeNanoseconds - start
     }
 


Arthur-Ficial (commented):

**Classification:** Code: non-network · `+16/-12` across 2 files · targets #436

- ⚠️ **Touches `Sources/**` with no test in this PR.** §7 requires a test for new behaviour on a code change. Worth weighing against the sibling candidates below.

### CI

### ⚠️ 3 competing candidates for #436

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#457 (this one)** | `+16/-12` | **no** | 🟢 green |
| #453 | `+130/-12` | yes | 🟢 green |
| #458 | `+137/-16` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:
Closing: #436 is fixed on `main` in 2ed07f6.

PR #458, opened 2026-09-06
fix(benchmark): make measure() rethrow so throwing ops fail the benchmark (#436)

### LINKED ISSUE #436: Benchmark cases discard errors with try?, so a failing operation reports as a fast validated sample

Arthur-Ficial:

Fixes #436.

## Summary

Three benchmark cases (`benchmarkContextManager`, `benchmarkRequestPipeline`, `benchmarkRequestDecode`) wrapped their timed operation in `try?`, so an operation that throws on every iteration was silently timed and reported as a fast, validated sample. The root cause is that `measure(iterations:)` accepted a non-throwing closure (`() async -> Void`), which forced every call site to discard errors at the boundary.

This makes `measure()` accept a throwing closure and `rethrow`. All three call sites now use `try` instead of `try?`. The `validated` flag is derived from the timed invocations (captured results from within the measure loop) rather than hardcoded to `true` or computed from a separate invocation.

## Root cause

`measure()` at `Sources/Benchmark.swift:433` took `() async -> Void`, so call sites had to silence throws with `try?`. Two cases (`benchmarkContextManager`, `benchmarkRequestDecode`) then set `validated: true` unconditionally, and the pipeline case (`benchmarkRequestPipeline`) validated a separate invocation rather than the one it timed. A benchmark measuring nothing looked like one that got faster.

## The fix

- `measure()` now takes `() async throws -> Void` and is marked `rethrows` - a throwing body propagates to the caller instead of being silently discarded.
- All three `try?` usages inside `measure` closures are replaced with `try`.
- `benchmarkContextManager`: captures the prompt from the last timed iteration for validation instead of hardcoding `validated: true`.
- `benchmarkRequestPipeline`: captures the result from the last timed iteration for validation instead of using a separate pre-loop invocation.
- `benchmarkRequestDecode`: captures the decoded request from the last timed iteration for validation instead of hardcoding `validated: true`. Function signature updated from `async` to `async throws`.

## Test coverage

- Added `Tests/apfelTests/BenchmarkTests.swift` with 5 tests covering the rethrow and validated-flag semantics:
  - `rethrowing measure propagates error from body on first iteration`
  - `rethrowing measure completes all iterations when body does not throw`
  - `rethrowing measure propagates error mid-sequence`
  - `a throwing body prevents reaching the validated assignment`
  - `non-throwing measure with try? body silently completes -- the bug pattern`
- Tests use local reimplementations of the old and new `measure()` signatures since the actual function is private in the FoundationModels-dependent target.

- All three `try?` usages inside `measure` closures removed - confirmed via grep.
- `measure()` signature changed to `rethrows` with throwing closure.
- `benchmarkRequestDecode` call site updated from `await` to `try await`.
- `validated` derived from timed invocations in all three cases.
- CHANGELOG.md `[Unreleased]` entry added.
- Diff limited to `Sources/Benchmark.swift`, `Tests/apfelTests/BenchmarkTests.swift`, `Tests/apfelTests/main.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.
- Swift 6 concurrency: mutable captures in `measure` closures are safe - `operation` is not `@Sendable` and runs sequentially. No data races introduced.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code. If the mutable variable captures in the `measure` closures cause Swift 6 strict concurrency diagnostics, the fix is to annotate them `nonisolated(unsafe)` (the accesses are sequential, no race).

Nothing - straightforward bug with a clear code-level fix. The issue body contained a suggested diff that matches the direction of this fix, but the implementation was written independently from reading the source.

## Test plan
- [ ] `swift run apfel-tests`
- [ ] `apfel --benchmark` produces correct output when nothing throws

---

https://claude.ai/code/session_01P5JVawtpoR2ZMofDzoh7So

---
_Generated by [Claude Code](https://claude.ai/code/session_01P5JVawtpoR2ZMofDzoh7So)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Benchmark cases no longer swallow errors via `try?` inside `measure()` closures (#436). A throwing timed operation now fails the benchmark instead of reporting a fast, validated sample. `measure()` accepts a throwing closure and rethrows; `validated` is derived from the timed invocations rather than hardcoded or computed from a separate call.
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Benchmark.swift
+++ b/Sources/Benchmark.swift
@@ -85,7 +85,7 @@ private func benchmarkReport() async throws -> BenchmarkReport {
     let streamDebugCapture = await benchmarkStreamDebugCaptureDisabled()
     let contextManager = try await benchmarkContextManager(options: options)
     let requestPipeline = try await benchmarkRequestPipeline(options: options)
-    let requestDecode = await benchmarkRequestDecode()
+    let requestDecode = try await benchmarkRequestDecode()
     let toolDetection = await benchmarkToolDetection()
     let responseEncode = await benchmarkResponseEncode()
 
@@ -239,14 +239,16 @@ private func benchmarkContextManager(options: SessionOptions) async throws -> Be
     let messages = benchmarkMessages()
 
     let iterations = 12
-    let timing = await measure(iterations: iterations) {
-        _ = try? await ContextManager.makeSession(
+    var lastPrompt = ""
+    let timing = try await measure(iterations: iterations) {
+        let (_, prompt, _) = try await ContextManager.makeSession(
             messages: messages,
             tools: tools,
             options: options,
             jsonMode: true,
             toolChoice: .auto
         )
+        lastPrompt = prompt
     }
 
     return BenchmarkCaseResult(
@@ -255,7 +257,7 @@ private func benchmarkContextManager(options: SessionOptions) async throws -> Be
         baseline_avg_ms: nil,
         current_avg_ms: timing.avgMilliseconds,
         speedup_ratio: nil,
-        validated: true,
+        validated: !lastPrompt.isEmpty,
         notes: "End-to-end session assembly with tools, JSON mode, and transcript trimming."
     )
 }
@@ -339,11 +341,11 @@ private func benchmarkStreamDebugCaptureDisabled() async -> BenchmarkCaseResult 
 private func benchmarkRequestPipeline(options: SessionOptions) async throws -> BenchmarkCaseResult {
     let requestJSON = makeRequestJSON()
     let request = try JSONDecoder().decode(ChatCompletionRequest.self, from: requestJSON)
-    let validation = try await benchmarkRequestPipelineResult(request: request, options: options)
 
     let iterations = 40
-    let timing = await measure(iterations: iterations) {
-        _ = try? await benchmarkRequestPipelineResult(request: request, options: options)
+    var lastResult: (finalPrompt: String, promptTokens: Int, responseBytes: Int) = ("", 0, 0)
+    let timing = try await measure(iterations: iterations) {
+        lastResult = try await benchmarkRequestPipelineResult(request: request, options: options)
     }
 
     return BenchmarkCaseResult(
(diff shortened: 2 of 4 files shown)


Arthur-Ficial:
CI status on `0682a36`:

| Check | Status | Notes |
|-------|--------|-------|
| `changelog-gate` | pass | `[Unreleased]` entry present |
| `build-and-test` | pass | Compilation clean, all unit tests + model-free integration tests green |
| `apfelcore-public-api` | fail | **Not this PR's failure** - see below |

The `apfelcore-public-api` job completed its actual API breakage check successfully ("No breaking changes detected in ApfelCore"). The job then failed in a subsequent `pip3 install` step because the runner's Homebrew-installed `typing_extensions 4.15.0` can't be uninstalled by pip (`error: uninstall-no-record-file`). This is a runner environment issue - the pip/Homebrew conflict exists regardless of this diff.

The fix would be adding `--ignore-installed` to the pip install command in `.github/workflows/ci.yml`, but that's CI infrastructure I'm not touching from this routine. A re-run would hit the same conflict.

The critical check (`build-and-test`) is green - the code compiles under Swift 6.3 strict concurrency and all tests pass.

cc @franzenzenhofer

---

Arthur-Ficial (commented):

**Classification:** Code: non-network · `+137/-16` across 4 files · targets #436

### CI

```
```

### ⚠️ 3 competing candidates for #436

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#458 (this one)** | `+137/-16` | yes | 🔴 pip flake only |
| #453 | `+130/-12` | yes | 🟢 green |
| #457 | `+16/-12` | **no** | 🟢 green |

Cheers, Arthur

Arthur-Ficial:
Closing: #436 is fixed on `main` in 2ed07f6.

**Not merged: a near-identical candidate was taken instead** (#457).

PR #460, opened 2026-09-06
fix(models): reject unrepresentable JSON numbers instead of silently rewriting to null (#455)

### LINKED ISSUE #455: AnyCodable rewrites unrepresentable JSON numbers to null instead of rejecting them

Arthur-Ficial:

Fixes #455.

## Summary

`AnyCodable.init(from:)` had an unconditional `value = nil` fallback at the end of its decode chain. Any JSON value that failed every decode attempt (in practice, a number outside `Double`'s range like `1e999`) was silently stored as `nil`, and `RawJSON` then re-encoded it as the JSON literal `null`. This meant schema constraints like `{"maximum": 1e999}` silently disappeared - the server accepted the request with HTTP 200 and applied a schema the caller never wrote.

The fix replaces the `value = nil` fallback with a `throw DecodingError.dataCorruptedError`, which the existing error handlers surface as HTTP 400 `invalid_request_error`. The explicit `container.decodeNil()` branch at the top of `init(from:)` already handles genuine JSON `null` values, so legitimate nulls are unaffected.

## Root cause

`Sources/Core/OpenAIModels.swift:431` - the terminal `value = nil` in `AnyCodable.init(from:)`. When `JSONDecoder` encounters a number outside `Double`'s representable range (e.g. `1e999`), the `try? container.decode(Double.self)` returns `nil`, and every other type attempt also fails, so the fallback fires and the value becomes `null`. `RawJSON.init(from:)` then re-encodes it, and the caller's schema constraint vanishes without any indication.

## The fix

One line changed in `Sources/Core/OpenAIModels.swift`: the `value = nil` fallback becomes a `throw DecodingError.dataCorruptedError(in:debugDescription:)`. Since `RawJSON.init(from:)` is already `throws`, and the HTTP handlers turn decode failures into 400 `invalid_request_error`, this is the minimal change needed.

## Test coverage

- Added `Tests/apfelTests/OpenAIModelsTests.swift`: `RawJSON rejects unrepresentable number (#455)` - asserts decoding `1e999` throws `DecodingError`
- Added `Tests/apfelTests/OpenAIModelsTests.swift`: `RawJSON preserves explicit null (#455)` - asserts a genuine JSON `null` still decodes correctly
- Added `Tests/apfelTests/OpenAIModelsTests.swift`: `RawJSON round-trips nested schema unchanged (#455)` - asserts a normal schema with numeric constraints round-trips intact
- Added `Tests/integration/server_validation_test.py`: `test_tool_parameters_with_unrepresentable_number_is_400` - asserts the server returns 400 for tool parameters containing `1e999`

## Test plan
- [ ] `swift run apfel-tests`

- Read and understood the `AnyCodable` decode/encode pair and the `RawJSON` re-encoding path
- Confirmed the existing `decodeNil()` branch handles genuine `null` before the fallback
- Confirmed `RawJSON.init(from:)` is `throws`, so the `DecodingError` propagates naturally
- Confirmed the HTTP handlers already turn `DecodingError` into 400 `invalid_request_error`
- Swift 6 concurrency: no data races introduced (no shared mutable state touched)
- Diff limited to `Sources/Core/OpenAIModels.swift`, `Tests/apfelTests/OpenAIModelsTests.swift`, `Tests/integration/server_validation_test.py`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies

## What I did NOT verify

Nothing - straightforward bug. The issue was filed by the repo owner and the reproduction is clean.

---

https://claude.ai/code/session_01K9wP7TW7G1AaHX21PCmUya

---
_Generated by [Claude Code](https://claude.ai/code/session_01K9wP7TW7G1AaHX21PCmUya)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `AnyCodable` no longer silently rewrites unrepresentable JSON numbers (e.g. `1e999`) to `null`. These values now produce a `DecodingError`, surfaced as HTTP 400 `invalid_request_error` on the server (#455).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -428,7 +428,9 @@ struct AnyCodable: Codable, Sendable {
             value = array
             return
         }
-        value = nil
+        throw DecodingError.dataCorruptedError(
+            in: container,
+            debugDescription: "Value is not representable as JSON (numbers must fit in a Double)")
     }
 
     func encode(to encoder: Encoder) throws {
--- a/Tests/apfelTests/OpenAIModelsTests.swift
+++ b/Tests/apfelTests/OpenAIModelsTests.swift
@@ -124,6 +124,31 @@ func runOpenAIModelsTests() {
         try assertEqual(parsed?[1] as? String, "two")
         try assertEqual(parsed?[2] as? Bool, false)
     }
+
+    test("RawJSON rejects unrepresentable number (#455)") {
+        do {
+            _ = try decode(RawJSON.self, from: "1e999")
+            try assertTrue(false, "should have thrown for unrepresentable number")
+        } catch is DecodingError {
+            // expected
+        }
+    }
+
+    test("RawJSON preserves explicit null (#455)") {
+        let raw = try decode(RawJSON.self, from: "null")
+        try assertEqual(raw.value, "null")
+    }
+
+    test("RawJSON round-trips nested schema unchanged (#455)") {
+        let input = #"{"type":"object","properties":{"x":{"type":"number","maximum":100}}}"#
+        let raw = try decode(RawJSON.self, from: input)
+        let reparsed = try JSONSerialization.jsonObject(with: Data(raw.value.utf8)) as? [String: Any]
+        let props = reparsed?["properties"] as? [String: Any]
+        let x = props?["x"] as? [String: Any]
+        try assertEqual(x?["maximum"] as? Int, 100)
+        try assertEqual(x?["type"] as? String, "number")
+        try assertEqual(reparsed?["type"] as? String, "object")
+    }
 }
 
 func runChatRequestValidatorTests() {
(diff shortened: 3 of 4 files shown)


Arthur-Ficial:
CI status update:

- `build-and-test` - passed
- `changelog-gate` - passed
- `apfelcore-public-api` - failed, but **not caused by this PR**

The `apfelcore-public-api` failure is in the `pip3 install` step, not the API breakage check. The actual API check passed ("No breaking changes detected in ApfelCore"). The runner's brew-managed `typing_extensions 4.15.0` now conflicts with pip trying to install `4.16.0` as a transitive dependency of `anyio`. This would reproduce on any branch - the last green CI on main ran Aug 5, before the runner environment changed.

The fix is in `ci.yml` (likely `pip3 install --break-system-packages --ignore-installed typing_extensions pytest httpx` or pinning the `anyio` version), but CI workflow changes are outside the scope of this routine. Flagging for Franz.

cc @franzenzenhofer

---

Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+57/-1` across 4 files · targets #455

### CI

```
```

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#460 (this one)** | `+57/-1` | yes | 🔴 pip flake only |
| #459 | `+54/-1` | yes | 🟢 green |
| #461 | `+88/-1` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:

**Not merged: a near-identical candidate was taken instead** (#459).

PR #461, opened 2026-09-06
fix(openai): reject unrepresentable JSON numbers instead of coercing to null (#455)

### LINKED ISSUE #455: AnyCodable rewrites unrepresentable JSON numbers to null instead of rejecting them

Arthur-Ficial:

Fixes #455.

## Root cause

`AnyCodable.init(from:)` has an unconditional `value = nil` fallback at its terminal branch. Any JSON value that fails every decode attempt (in practice, a number outside `Double`'s representable range like `1e999`) is silently stored as `nil`. When `RawJSON.init(from:)` re-encodes it, `nil` becomes the JSON literal `null`. The caller's schema constraint simply disappears - a `{"maximum": 1e999}` becomes `{"maximum": null}` - and the server returns HTTP 200 with a schema that no longer matches what was sent.

The explicit `container.decodeNil()` branch at line 418 already handles genuine JSON `null` values, so the terminal fallback only fires for truly unrepresentable values.

## The fix

Replace the `value = nil` fallback with a `throw DecodingError.dataCorruptedError(in:debugDescription:)`. The existing handler in `Handlers.swift` already catches `DecodingError` from `JSONDecoder().decode(ChatCompletionRequest.self, from: body)` and maps it to HTTP 400 `invalid_request_error`, so no plumbing changes are needed. The error message reads: "Value is not representable as JSON (numbers must fit in a 64-bit float)".

One-line change in production code. The `default: try container.encodeNil()` arm in `encode(to:)` becomes unreachable but is harmless to leave.

## Test coverage

- Added `OpenAIModelsTests.swift`: `AnyCodable rejects unrepresentable number (1e999) instead of coercing to null` - asserts decoding `1e999` via `RawJSON` throws `DecodingError`.
- Added `OpenAIModelsTests.swift`: `AnyCodable rejects negative overflow number (-1e999)` - same for negative overflow.
- Added `OpenAIModelsTests.swift`: `AnyCodable preserves explicit JSON null` - asserts a genuine `null` still decodes as `"null"`.
- Added `OpenAIModelsTests.swift`: `AnyCodable round-trips nested schema unchanged` - asserts a normal schema with `{"maximum": 100}` survives decode/re-encode.
- Added `OpenAIModelsTests.swift`: `Tool parameters with nested unrepresentable number reject at decode` - asserts a full `OpenAITool` with nested `1e999` in parameters throws.
- Added `server_validation_test.py`: `test_tool_parameters_with_unrepresentable_number_is_400` - integration test confirming the HTTP surface returns 400 for `float("inf")` in tool parameters.
- Existing `RawJSON preserves nested tool parameter schemas as valid JSON` still covers the happy path.

- Read `AnyCodable.init(from:)` and confirmed the `value = nil` fallback at line 431 is the only code path that can produce a silent null substitution.
- Read `RawJSON.init(from:)` at lines 286-291 and confirmed it decodes via `AnyCodable` and re-encodes, propagating the null.
- Read `Handlers.swift` lines 64-75 and confirmed `JSONDecoder().decode(ChatCompletionRequest.self, from: body)` is wrapped in do/catch that maps any decode error to 400 `invalid_request_error`.
- Confirmed `container.decodeNil()` at line 418 handles genuine JSON nulls before the fallback, so the change does not affect legitimate null values.
- Swift 6 concurrency: no data races introduced (pure synchronous decode path, no shared mutable state).
- Diff limited to `Sources/Core/OpenAIModels.swift`, `Tests/apfelTests/OpenAIModelsTests.swift`, `Tests/integration/server_validation_test.py`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code or `swift build` on this Linux runner. If the fix does not actually resolve the reported behaviour, that is on me to refine - ping me in a review comment.

Nothing - straightforward bug with a clear root cause and minimal fix. The issue was filed by `Arthur-Ficial` (owner account) and the analysis is accurate.

---

https://claude.ai/code/session_01U6VhrGyeLFhcyTMySjg87i

---
_Generated by [Claude Code](https://claude.ai/code/session_01U6VhrGyeLFhcyTMySjg87i)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `AnyCodable` no longer silently coerces unrepresentable JSON numbers (e.g. `1e999`) to `null`. Tool parameter schemas and `response_format.json_schema.schema` payloads containing such values now return HTTP 400 `invalid_request_error` instead of silently rewriting the caller's JSON (#455).
+
 ## [1.9.1] - 2026-08-05
 
 ### Changed
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -428,7 +428,9 @@ struct AnyCodable: Codable, Sendable {
             value = array
             return
         }
-        value = nil
+        throw DecodingError.dataCorruptedError(
+            in: container,
+            debugDescription: "Value is not representable as JSON (numbers must fit in a 64-bit float)")
     }
 
     func encode(to encoder: Encoder) throws {
--- a/Tests/apfelTests/OpenAIModelsTests.swift
+++ b/Tests/apfelTests/OpenAIModelsTests.swift
@@ -124,6 +124,57 @@ func runOpenAIModelsTests() {
         try assertEqual(parsed?[1] as? String, "two")
         try assertEqual(parsed?[2] as? Bool, false)
     }
+
+    // MARK: - #455 AnyCodable rejects unrepresentable values
+
+    test("AnyCodable rejects unrepresentable number (1e999) instead of coercing to null (#455)") {
+        do {
+            _ = try decode(RawJSON.self, from: "1e999")
+            throw TestFailure("expected DecodingError for 1e999, none thrown")
+        } catch is TestFailure {
+            throw TestFailure("expected DecodingError for 1e999, none thrown")
+        } catch {
+            try assertTrue(error is DecodingError, "expected DecodingError, got \(type(of: error))")
+        }
+    }
+
+    test("AnyCodable rejects negative overflow number (-1e999) (#455)") {
+        do {
+            _ = try decode(RawJSON.self, from: "-1e999")
+            throw TestFailure("expected DecodingError for -1e999, none thrown")
+        } catch is TestFailure {
+            throw TestFailure("expected DecodingError for -1e999, none thrown")
+        } catch {
+            try assertTrue(error is DecodingError, "expected DecodingError, got \(type(of: error))")
+        }
+    }
+
+    test("AnyCodable preserves explicit JSON null (#455)") {
+        let raw = try decode(RawJSON.self, from: "null")
+        try assertEqual(raw.value, "null")
+    }
+
+    test("AnyCodable round-trips nested schema unchanged (#455)") {
+        let input = #"{"type":"object","properties":{"x":{"type":"number","maximum":100}}}"#
+        let raw = try decode(RawJSON.self, from: input)
+        let reparsed = try JSONSerialization.jsonObject(with: Data(raw.value.utf8)) as? [String: Any]
+        let props = reparsed?["properties"] as? [String: Any]
+        let x = props?["x"] as? [String: Any]
+        try assertEqual(x?["maximum"] as? Int, 100)
+        try assertEqual(x?["type"] as? String, "number")
+    }
+
+    test("Tool parameters with nested unrepresentable number reject at decode (#455)") {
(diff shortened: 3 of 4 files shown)


Arthur-Ficial:
**CI status update**

- `changelog-gate` - passed
- `build-and-test` - still queued
- `apfelcore-public-api` - failed, but **not this PR's fault**

The `apfelcore-public-api` check completed its actual job successfully - "No breaking changes detected in ApfelCore" / "No non-additive API breakage." The failure is in a subsequent `pip3 install --break-system-packages pytest httpx` step: Homebrew-installed `typing_extensions` 4.15.0 has no RECORD file, so pip can't upgrade it to 4.16.0 (required by `anyio`). This is a runner environment issue - the last 5 CI runs on main all passed (most recent: 2026-08-05), so the runner image has likely drifted since then.

I don't have the means to re-run the check from here. Franz can re-run it, or the fix is to add `--ignore-installed typing_extensions` to the pip command in the workflow (but that's a CI change, which I don't touch from a routine).

Will keep watching for `build-and-test` to complete.

---

Arthur-Ficial:
**CI update - `build-and-test` also hit the same runner issue**

`build-and-test` completed its Swift work successfully:
- Release build: clean, no warnings
- Unit tests: **all 1047 passed** (1042 existing + 5 new from this PR)

Then it failed on the identical `pip3 install` step - same Homebrew `typing_extensions` 4.15.0 vs pip 4.16.0 conflict. Both checks (`apfelcore-public-api` and `build-and-test`) are blocked by this runner environment issue, not by anything in this PR's diff.

**Summary of CI:**
| Check | Swift work | pip install | Overall |
|-------|-----------|-------------|---------|
| `changelog-gate` | n/a | n/a | passed |
| `apfelcore-public-api` | passed (no API breakage) | failed (typing_extensions) | failed |
| `build-and-test` | passed (1047/1047 tests) | failed (typing_extensions) | failed |

I can't re-run these checks from here. Franz, when you pick this up, a re-run might work if the runner image has been updated, or the CI workflow's pip install step needs `--ignore-installed typing_extensions` added.

---

Arthur-Ficial (commented):

**Classification:** Code: network/parsing/auth · `+88/-1` across 4 files · targets #455

### CI
Red on: `apfelcore-public-api`, `build-and-test` — but **this is not your diff.**

```
```

| PR | size | tests | CI | 
| --- | --- | --- | --- |
| **#461 (this one)** | `+88/-1` | yes | 🔴 pip flake only |
| #459 | `+54/-1` | yes | 🟢 green |
| #460 | `+57/-1` | yes | 🔴 pip flake only |

Cheers, Arthur

Arthur-Ficial:

**Not merged: a near-identical candidate was taken instead** (#459).

PR #488, opened 2026-09-07
fix(server): honor max_completion_tokens from modern OpenAI clients (#478)

### LINKED ISSUE #478: fix: Honor max_completion_tokens from current OpenAI-compatible clients

Arthur-Ficial:

Fixes #478.

## Summary

`ChatCompletionRequest` only decoded `max_tokens`, silently ignoring the modern `max_completion_tokens` field that LangChain and current openai-python send. The caller's output-token cap was dropped and the model generated an unbounded response.

Both fields are now decoded and resolved to one effective limit via `effectiveMaxTokens`. Providing both with different values returns HTTP 400 with `param: "max_completion_tokens"`. `/v1/models` advertises the new field in `supported_parameters`.

## Root cause

`ChatCompletionRequest` (Sources/Core/OpenAIModels.swift:23) declared only `max_tokens`. Unknown JSON keys are silently dropped by Swift's `Decodable` synthesis. `Handlers.swift:143` forwarded only `chatRequest.max_tokens` into `SessionOptions`, and `ChatRequestValidator.swift:218` validated only it. LangChain rewrites even a legacy `max_tokens` to `max_completion_tokens` at send time, so the limit was always lost for those clients.

## The fix

- Added `max_completion_tokens: Int?` property to `ChatCompletionRequest` with a new `effectiveMaxTokens` computed property that prefers the modern field, falling back to legacy.
- Added `max_completion_tokens <= 0` validation and a new `conflictingMaxTokens` failure case in `ChatRequestValidator` for when both fields are present with different values (HTTP 400, `param: "max_completion_tokens"`).
- Changed `Handlers.swift` to pass `chatRequest.effectiveMaxTokens` instead of `chatRequest.max_tokens` into `SessionOptions`.
- Added `max_completion_tokens` to `/v1/models` `supported_parameters`.
- Extended `ApfelCorePublicAPIUsageTests` to cover the new property and failure case.

## Test plan

- [ ] `swift run apfel-tests`

## Test coverage

- Added 16 tests in `OpenAIModelsTests.swift`:
  - Decoding: `max_completion_tokens` present, absent, both fields together
  - `effectiveMaxTokens` resolution: modern-only, legacy-only, both equal, neither
  - Init round-trip for the new field
  - Validator: rejects zero, rejects negative, accepts positive, rejects conflict, accepts identical pair, accepts either alone
  - Conflict metadata: httpStatusCode=400, errorParam, message content
  - Validation ordering: invalid individual values reported before conflict
  - LangChain request shape fixture
  - openai-python request shape fixture (streaming)
- `ApfelCorePublicAPIUsageTests` updated for `max_completion_tokens`, `effectiveMaxTokens`, and `conflictingMaxTokens`.
- Existing `max_tokens` tests unchanged and still pass.

- Diff limited to `OpenAIModels.swift`, `ChatRequestValidator.swift`, `Handlers.swift`, `Server.swift`, `OpenAIModelsTests.swift`, `ApfelCorePublicAPIUsageTests.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.
- `effectiveMaxTokens` correctly prefers `max_completion_tokens` over `max_tokens` (modern wins).
- Conflict detection only fires after individual range checks, so `max_tokens: 0, max_completion_tokens: 100` reports the zero, not the conflict.
- Public init has `max_completion_tokens` as a defaulted parameter after `max_tokens`, preserving backward compatibility for all existing callers.
- `CHANGELOG.md` has an `[Unreleased]` entry (changelog-gate CI requirement).

## What I did NOT verify

Nothing - straightforward bug. The issue was filed by the repo owner with clear references to upstream LangChain and OpenAI API docs.

---

https://claude.ai/code/session_0154mjXU4Ghpqg32JcGz6nVP

---
_Generated by [Claude Code](https://claude.ai/code/session_0154mjXU4Ghpqg32JcGz6nVP)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `max_completion_tokens` in chat-completions requests was silently ignored because `ChatCompletionRequest` only decoded `max_tokens`. Modern OpenAI clients (LangChain, openai-python) send `max_completion_tokens` instead, so the caller's output-token cap was dropped and the model generated an unbounded response. Both fields are now decoded and resolved to one effective limit; providing both with different values returns 400 with `param: "max_completion_tokens"`. `/v1/models` advertises the new field (#478).
+
 ## [1.10.0] - 2026-09-07
 
 ### Fixed
--- a/Sources/Core/ChatRequestValidator.swift
+++ b/Sources/Core/ChatRequestValidator.swift
@@ -83,6 +83,9 @@ public enum ChatRequestValidationFailure: Sendable, Equatable, Hashable, CustomS
     case imageContent
     /// A numeric or string parameter had an invalid value.
     case invalidParameterValue(String)
+    /// Both `max_tokens` and `max_completion_tokens` were provided with
+    /// different values.
+    case conflictingMaxTokens(legacy: Int, modern: Int)
     /// The request asked for a model name other than `apple-foundationmodel`.
     case invalidModel(String)
 
@@ -103,6 +106,8 @@ public enum ChatRequestValidationFailure: Sendable, Equatable, Hashable, CustomS
             return "Image content is not supported by the Apple on-device model"
         case .invalidParameterValue(let detail):
             return detail
+        case .conflictingMaxTokens(let legacy, let modern):
+            return "Both 'max_tokens' (\(legacy)) and 'max_completion_tokens' (\(modern)) were provided with different values. Use 'max_completion_tokens' only, or provide the same value for both."
         case .invalidModel(let model):
             return "The model '\(model)' does not exist. The only available model is 'apple-foundationmodel'."
         }
@@ -125,6 +130,8 @@ public enum ChatRequestValidationFailure: Sendable, Equatable, Hashable, CustomS
             return "rejected: image content"
         case .invalidParameterValue(let detail):
             return "validation failed: \(detail)"
+        case .conflictingMaxTokens(let legacy, let modern):
+            return "validation failed: conflicting max_tokens=\(legacy) vs max_completion_tokens=\(modern)"
         case .invalidModel(let model):
             return "validation failed: unknown model \(model)"
         }
@@ -158,6 +165,8 @@ public enum ChatRequestValidationFailure: Sendable, Equatable, Hashable, CustomS
             return "model"
         case .unknownRole:
             return "messages"
+        case .conflictingMaxTokens:
+            return "max_completion_tokens"
         default:
             return nil
         }
@@ -217,6 +226,12 @@ public enum ChatRequestValidator {
 
         if let maxTokens = request.max_tokens, maxTokens <= 0 {
             return .invalidParameterValue("'max_tokens' must be a positive integer, got \(maxTokens)")
+        }
+        if let maxCompletionTokens = request.max_completion_tokens, maxCompletionTokens <= 0 {
+            return .invalidParameterValue("'max_completion_tokens' must be a positive integer, got \(maxCompletionTokens)")
+        }
+        if let legacy = request.max_tokens, let modern = request.max_completion_tokens, legacy != modern {
+            return .conflictingMaxTokens(legacy: legacy, modern: modern)
         }
         if let temp = request.temperature, temp < 0 {
             return .invalidParameterValue("'temperature' must be non-negative, got \(temp)")
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -52,6 +52,8 @@ public struct ChatCompletionRequest: Decodable, Sendable, Equatable, Hashable {
 
     /// The effective positive output-token limit after resolving `max_tokens`
     /// and `max_completion_tokens`. `nil` when neither field was provided.
+    /// On the server path the validator rejects conflicting values, so the
+    /// precedence matters only for direct `ApfelCore` library consumers.
     public var effectiveMaxTokens: Int? {
         max_completion_tokens ?? max_tokens
     }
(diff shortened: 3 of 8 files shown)


Arthur-Ficial (commented):

**Headline: this and #487 are duplicate fixes for #478** — opened 34 seconds apart by two separate bug-solver sessions. Only one can merge. **Of the two, this is the one I would keep** (it advertises the field in `/v1/models`, sets a real `error.param` via a typed failure case, and delegates rather than duplicating the initializer body — full comparison in my review on #487). Franz makes the call; I have closed nothing.

### Correctness — looks right

- `effectiveMaxTokens = max_completion_tokens ?? max_tokens` is the correct precedence per the OpenAI vocabulary, and `Handlers.swift` now feeds it into `SessionOptions.maxTokens`, which is the single path into `GenerationOptions.maximumResponseTokens`.
- Validator ordering is deliberate and correct: both individual `<= 0` checks fire before the conflict guard, so `{"max_tokens": 0, "max_completion_tokens": 100}` reports the zero rather than the conflict. The tests cover this ordering explicitly.
- The legacy initializer overload delegates to the designated one — no duplicated assignment list to drift.
- `conflictingMaxTokens` is safe to add: STABILITY.md states public `ApfelCore` enums are non-frozen and that the API-breakage gate "permits added enum cases". The `apfelcore-public-api` check is green, which confirms it.
- `CLIServerParityTests` was updated to assert the *new* invariant (`maxTokens: chatRequest.effectiveMaxTokens,`) rather than deleted — the guardrail is preserved, not weakened. Good.

### Two things worth flagging before merge

**1. Merge-order hazard with #489 — this one matters.** Both this PR and #489 add a *backwards-compatible initializer overload with the identical 20-parameter legacy signature* (`model:messages:stream:stream_options:temperature:top_p:max_tokens:seed:tools:tool_choice:response_format:logprobs:n:stop:presence_penalty:frequency_penalty:user:x_context_strategy:x_context_max_turns:x_context_output_reserve:`). They will conflict in git, and if the conflict is resolved by keeping both, the result is an *invalid redeclaration* — it will not compile. Whichever merges second must fold its new field into the *other's* designated initializer and delete its own overload, not add a second one. Please rebase and re-run CI on the second PR rather than trusting the current green.

**2. The "modern wins" precedence is unreachable on the server path.** Because the validator rejects any request where both fields are present with different values, `effectiveMaxTokens` can only ever be choosing between "equal values" or "one is nil". The precedence is therefore live only for direct `ApfelCore` library consumers, who construct `ChatCompletionRequest` without going through `ChatRequestValidator`. That is not a bug and I would not change the code — but the doc comment reads like a server-side rule, and it is not one. A half-line clarifying that would save a future reader the same trip.

### On verification

The PR body's "I cannot run FoundationModels code" is narrower than it reads: CI runs on `macos-26` and **this PR is green** — `build-and-test` (`make build` + `swift run apfel-tests` + the model-free integration and HTTP server suites), `apfelcore-public-api` (including `swift package diagnose-api-breaking-changes` and the strict-concurrency build), and `changelog-gate`. The 16 new unit tests do run and do pass. What remains genuinely unproven is the live end-to-end behaviour — that `max_completion_tokens: 64` actually produces a capped generation from the real on-device model. That needs @franzenzenhofer locally, and the `Test plan` checkboxes in the description are still unticked.

Cheers, Arthur

Issue #238, opened 2026-07-01
fix(sse): non-usage chunks omit "usage": null with include_usage; invalid tool_choice silently coerced to auto


Arthur-Ficial:
## Summary
Two OpenAI wire-format deviations:
1. With `stream_options.include_usage: true`, OpenAI sends `"usage": null` on every chunk except the final stats chunk; apfel omits the key entirely (`Sources/Models.swift:97-103`, synthesized encoder drops optional `usage`).
2. `ToolChoice.init(from:)` (`Sources/Core/OpenAIModels.swift:337-365`) maps any unrecognized string (`"banana"`) and any undecodable object to `.auto`; OpenAI rejects invalid `tool_choice` with 400.

## Suggested fix
- Add a manual `encode(to:)` on `ChatCompletionChunk` that emits explicit `usage: null` when the flag is set (plumb `includeUsage` through the SSE builders).
- Make `ToolChoice` decoding throw (or decode to `.invalid(String)` rejected by the validator).

## Severity
Low.

---
_Filed from a full read-only audit of apfel v1.6.1 (server, CLI/chat, MCP/tool-calling, security, concurrency, tests/CI/docs). Findings verified against committed code; high-severity crash/DoS items were reproduced against the release binary. One issue per finding._

Arthur-Ficial:
Fixed in 2efc149 - with stream_options.include_usage every non-final chunk now carries explicit usage:null, and invalid tool_choice values are rejected with 400 instead of being coerced to auto. Ships in the next release.

PR #489, opened 2026-09-07
fix(server): decode parallel_tool_calls, validate named tool_choice (#480)

### LINKED ISSUE #480: fix: Enforce tool_choice and parallel_tool_calls at the response boundary
**Value:** Agent clients use tool controls to decide whether to execute a function, wait for required arguments, or handle an ordinary answer. These controls must be enforceable contracts for tools that inspect a repository or change local state. A user should also be able to ask apfel to explain tool-call JSON without having that explanation converted into an executable request.

**Problem today:** [Sources/ContextManager.swift:138](https://github.com/Arthur-Ficial/apfel/blob/f8942de1661370a5c125ef8f494cd3d65eaaaba0/Sources/ContextManager.swift#L138) implements valid choices through instructions; `.specific` still presents all effective tool definitions. Validation at `Sources/Core/ChatRequestValidator.swift:243` catches malformed choices but does not verify required/selected tools exist. [Sources/Handlers.swift:456](https://github.com/Arthur-Ficial/apfel/blob/f8942de1661370a5c125ef8f494cd3d65eaaaba0/Sources/Handlers.swift#L456) and `:587` detect tool-call-shaped output without checking an effective allowed-tool policy. Thus prose steering is the last barrier against an unwanted control response, even when no tools should be active. `Sources/Core/OpenAIModels.swift:26` has no `parallel_tool_calls` property. #238 fixed invalid choice syntax; #224 fixed chunk framing; this proposal concerns the semantics of valid requests.

**Proposal:** Resolve a pure request-scoped `ToolPolicy` containing allowed names, required-call behavior, and maximum call count. `.none` disables both client and injected MCP tool dispatch, and tool-shaped JSON is ordinary content when the effective tool scope is empty. A named choice restricts generation and validation to that name; `.required` requires a valid allowed call unless the model returns an explicit refusal. Decode `parallel_tool_calls`; `false` permits at most one returned call.

Use guided generation for forced-call envelopes where the supported parameter schema allows it, while retaining the existing out-of-band execution architecture. Validate every returned call before exposing or dispatching it. Unsatisfied model output becomes a typed `tool_choice_not_satisfied` error, not HTTP 200 with an apparently completed answer; any regeneration must be bounded and must never re-execute tools. For `request.json` containing a function definition named `lookup_ticket`, a client can request one specific call with:

```bash
jq '.tool_choice={"type":"function","function":{"name":"lookup_ticket"}} | .parallel_tool_calls=false' request.json \
  | curl --fail-with-body http://localhost:11434/v1/chat/completions \
      -H 'Content-Type: application/json' --data-binary @-
```

**Acceptance criteria:**

- Begin with deterministic failing policy tests: no tools, `.none`, `.auto`, `.required`, a valid named choice, an unknown name, and `parallel_tool_calls:false` with two generated calls.
- Invalid request combinations return HTTP 400 with the offending parameter; model contract failures return a typed server error, with correct SSE error termination when headers were already sent.
- Tool-call examples in ordinary content, including `json_object` output, remain content when tools are disabled; no MCP fixture invocation occurs.
- Every emitted or executed function name belongs to the resolved allowed set; forced names and single-call limits hold in streaming and non-streaming paths.
- Preserve explicit refusal behavior and existing framing tests; add a real SDK forced-tool round trip and deterministic failure coverage that does not depend on the model randomly disobeying.

**Golden-goal alignment:** Makes the OpenAI server's tool-calling contract reliable enough for real client-controlled workflows.

**Effort:** L

**Risks / open questions:**

- A forced schema can ensure structure, not useful reasoning; model refusals remain legitimate outcomes.
- Bounded repair adds latency; prefer a guided single call where possible and measure extra attempts before choosing defaults.
- Keep FoundationModels live `Tool` registration out of this change; adopting it would trigger the separate #119/#197 architectural requirements.

Arthur-Ficial:

Refs #480 (step 1 of a scope-split - see issue comment for the full plan).

## Summary

Two request-level bugs in the OpenAI-compatible chat completions endpoint:

1. **`parallel_tool_calls` silently ignored.** The field was absent from `ChatCompletionRequest`, so `{"parallel_tool_calls": false}` decoded without error but the value was discarded. Now decoded as `Bool?`.

2. **Named `tool_choice` accepted impossible names.** A request with `tool_choice: {"type":"function","function":{"name":"nonexistent"}}` and a `tools` array that did not contain `nonexistent` passed validation and reached the model with a prompt instruction to call a function that was never defined. Now returns 400 with the offending name and the available names.

## Root cause

`ChatCompletionRequest` had no `parallel_tool_calls` property, so Swift's synthesized `Decodable` silently skipped the key. `ChatRequestValidator.validate` checked for `.invalid` format (malformed string/object, #238) but never compared a `.specific(name:)` against the `tools` array.

## The fix

- Added `parallel_tool_calls: Bool?` to `ChatCompletionRequest` (property, init parameter, init body).
- Added a backwards-compatible init overload preserving the pre-#480 signature so `swift package diagnose-api-breaking-changes` passes (the old `init(model:messages:...:tool_choice:response_format:...)` symbol must stay in the API dump).
- After the existing `.invalid` tool_choice check in `ChatRequestValidator`, added a guard: when `tool_choice` is `.specific(name:)` AND `tools` is non-empty, the name must appear in the tools array. When `tools` is nil or empty, the check is skipped because MCP-injected tools resolve after validation (documented in a code comment at the guard).
- Updated the public API surface test.

## Test coverage

- Added `OpenAIModelsTests`: 3 tests for `parallel_tool_calls` decoding (true, false, nil-when-absent).
- Added `OpenAIModelsTests`: 4 tests for named tool_choice validation (name not in tools, name matches, empty tools array skipped, multiple tools none match).
- Existing `tool_choice` format tests (#238) unchanged - they send `tool_choice` without tools, which remains valid.

- Diff limited to `OpenAIModels.swift`, `ChatRequestValidator.swift`, `OpenAIModelsTests.swift`, `ApfelCorePublicAPIUsageTests.swift`, `CHANGELOG.md`.
- No touches to release infrastructure, guardrails, CI, or dependencies.
- Swift 6 concurrency: all new types are `Sendable`-compatible (existing `Bool?` property on an existing `Sendable` struct).
- Existing integration tests all provide matching tool names when using specific `tool_choice`, so no breakage expected.

## What I did NOT verify

Nothing - straightforward validation gap. The issue was filed by the project owner with detailed source references.

## Remaining work (#480 scope-split)

This PR is step 1. The full issue requires:

- **Step 2: Response-boundary enforcement** - Gate `ToolCallHandler.detectToolCall()` on `tool_choice` in `Handlers.swift` (non-streaming and streaming paths). When `tool_choice: none`, tool-call-shaped output becomes ordinary content. This step should also close the MCP gap where a named choice can still reference a tool no MCP server provides.
- **Step 3: `parallel_tool_calls: false` enforcement** - Limit detected tool calls to one when the flag is false.
- **Step 4: Typed errors** - `tool_choice_not_satisfied` error type for when the model ignores a forced choice. SSE error termination for mid-stream contract violations.
- **Step 5: `supported_parameters` + close** - Add `parallel_tool_calls` to the `/v1/models` response's `supported_parameters` list in `Server.swift`. Close #480.

## Merge-order note

#488 adds a backwards-compatible init overload with the same 20-parameter legacy signature. Whichever merges second must fold its new field into the other's designated init and resolve the conflict. See review thread for details.

## Test plan

- [ ] `swift run apfel-tests`

---

https://claude.ai/code/session_0199eW8kcWiThZVk5xjjKaxW

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [https://semver.org/](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `ChatCompletionRequest` now decodes `parallel_tool_calls` (previously silently ignored). A named `tool_choice` referencing a function not present in the `tools` array is now rejected with a 400 instead of reaching the model with an impossible constraint (#480).
+
 ## [1.10.0] - 2026-09-07
 
 ### Fixed
--- a/Sources/Core/ChatRequestValidator.swift
+++ b/Sources/Core/ChatRequestValidator.swift
@@ -243,6 +243,12 @@ public enum ChatRequestValidator {
         if case .invalid(let raw) = request.tool_choice {
             return .invalidParameterValue("Invalid 'tool_choice' value: \(raw). Must be 'auto', 'none', 'required', or a {\"type\":\"function\",\"function\":{\"name\":\"...\"}} object.")
         }
+        if case .specific(let name) = request.tool_choice,
+           let tools = request.tools, !tools.isEmpty {
+            if !tools.contains(where: { $0.function.name == name }) {
+                return .invalidParameterValue("tool_choice function name '\(name)' does not match any tool in the 'tools' array. Available: \(tools.map(\.function.name).joined(separator: ", "))")
+            }
+        }
 
         return nil
     }
--- a/Sources/Core/ChatRequestValidator.swift
+++ b/Sources/Core/ChatRequestValidator.swift
@@ -243,6 +243,8 @@ public enum ChatRequestValidator {
         if case .invalid(let raw) = request.tool_choice {
             return .invalidParameterValue("Invalid 'tool_choice' value: \(raw). Must be 'auto', 'none', 'required', or a {\"type\":\"function\",\"function\":{\"name\":\"...\"}} object.")
         }
+        // Skipped when tools is nil/empty: MCP-injected tools resolve after
+        // validation, so a named choice without client tools is not yet invalid.
         if case .specific(let name) = request.tool_choice,
            let tools = request.tools, !tools.isEmpty {
             if !tools.contains(where: { $0.function.name == name }) {
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -96,6 +96,44 @@ public struct ChatCompletionRequest: Decodable, Sendable, Equatable, Hashable {
         self.x_context_max_turns = x_context_max_turns
         self.x_context_output_reserve = x_context_output_reserve
     }
+
+    /// Backwards-compatible overload preserving the pre-#480 init signature.
+    public init(
+        model: String,
+        messages: [OpenAIMessage],
+        stream: Bool? = nil,
+        stream_options: StreamOptions? = nil,
+        temperature: Double? = nil,
+        top_p: Double? = nil,
+        max_tokens: Int? = nil,
+        seed: Int? = nil,
+        tools: [OpenAITool]? = nil,
+        tool_choice: ToolChoice? = nil,
+        response_format: ResponseFormat? = nil,
+        logprobs: Bool? = nil,
+        n: Int? = nil,
+        stop: RawJSON? = nil,
+        presence_penalty: Double? = nil,
+        frequency_penalty: Double? = nil,
+        user: String? = nil,
+        x_context_strategy: String? = nil,
+        x_context_max_turns: Int? = nil,
+        x_context_output_reserve: Int? = nil
+    ) {
+        self.init(
+            model: model, messages: messages, stream: stream,
+            stream_options: stream_options, temperature: temperature,
+            top_p: top_p, max_tokens: max_tokens, seed: seed,
+            tools: tools, tool_choice: tool_choice,
+            parallel_tool_calls: nil,
+            response_format: response_format, logprobs: logprobs,
+            n: n, stop: stop, presence_penalty: presence_penalty,
+            frequency_penalty: frequency_penalty, user: user,
+            x_context_strategy: x_context_strategy,
+            x_context_max_turns: x_context_max_turns,
+            x_context_output_reserve: x_context_output_reserve
+        )
+    }
 }
 
 /// OpenAI `stream_options` payload.
--- a/Sources/Core/OpenAIModels.swift
+++ b/Sources/Core/OpenAIModels.swift
@@ -27,6 +27,8 @@ public struct ChatCompletionRequest: Decodable, Sendable, Equatable, Hashable {
     public let tools: [OpenAITool]?
     /// How the client wants tool choice resolved.
     public let tool_choice: ToolChoice?
+    /// Whether the model may return multiple tool calls in one response.
+    public let parallel_tool_calls: Bool?
     /// Requested response-format contract.
     public let response_format: ResponseFormat?
     /// OpenAI logprobs request flag.
@@ -60,6 +62,7 @@ public struct ChatCompletionRequest: Decodable, Sendable, Equatable, Hashable {
         seed: Int? = nil,
         tools: [OpenAITool]? = nil,
         tool_choice: ToolChoice? = nil,
+        parallel_tool_calls: Bool? = nil,
         response_format: ResponseFormat? = nil,
         logprobs: Bool? = nil,
         n: Int? = nil,
@@ -81,6 +84,7 @@ public struct ChatCompletionRequest: Decodable, Sendable, Equatable, Hashable {
         self.seed = seed
         self.tools = tools
         self.tool_choice = tool_choice
+        self.parallel_tool_calls = parallel_tool_calls
         self.response_format = response_format
         self.logprobs = logprobs
         self.n = n
--- a/Tests/apfelTests/ApfelCorePublicAPIUsageTests.swift
+++ b/Tests/apfelTests/ApfelCorePublicAPIUsageTests.swift
@@ -214,6 +214,7 @@ func runApfelCorePublicAPIUsageTests() {
         let _: Int?                = req.seed
         let _: [OpenAITool]?       = req.tools
         let _: ToolChoice?         = req.tool_choice
+        let _: Bool?               = req.parallel_tool_calls
         let _: ResponseFormat?     = req.response_format
         let _: Bool?               = req.logprobs
         let _: Int?                = req.n
--- a/Tests/apfelTests/OpenAIModelsTests.swift
+++ b/Tests/apfelTests/OpenAIModelsTests.swift
@@ -682,6 +682,69 @@ func runChatRequestValidatorTests() {
         )
         try assertEqual(ChatRequestValidator.validate(request), .unknownRole("bogus"))
     }
+
+    // --- parallel_tool_calls decoding (#480) ---
+
+    test("ChatCompletionRequest decodes parallel_tool_calls=true (#480)") {
+        let json = #"{"model":"apple-foundationmodel","messages":[{"role":"user","content":"hi"}],"parallel_tool_calls":true}"#
+        let req = try decode(ChatCompletionRequest.self, from: json)
+        try assertEqual(req.parallel_tool_calls, true)
+    }
+
+    test("ChatCompletionRequest decodes parallel_tool_calls=false (#480)") {
+        let json = #"{"model":"apple-foundationmodel","messages":[{"role":"user","content":"hi"}],"parallel_tool_calls":false}"#
+        let req = try decode(ChatCompletionRequest.self, from: json)
+        try assertEqual(req.parallel_tool_calls, false)
+    }
+
+    test("ChatCompletionRequest parallel_tool_calls is nil when absent (#480)") {
+        let json = #"{"model":"apple-foundationmodel","messages":[{"role":"user","content":"hi"}]}"#
+        let req = try decode(ChatCompletionRequest.self, from: json)
+        try assertNil(req.parallel_tool_calls)
+    }
+
+    // --- Named tool_choice validation against tools array (#480) ---
+
+    test("validator rejects named tool_choice when name not in tools (#480)") {
+        let request = try decode(
+            ChatCompletionRequest.self,
+            from: #"{"model":"\#(M)","messages":[{"role":"user","content":"hi"}],"tools":[{"type":"function","function":{"name":"bar","description":"d"}}],"tool_choice":{"type":"function","function":{"name":"foo"}}}"#
+        )
+        guard case .invalidParameterValue(let detail) = ChatRequestValidator.validate(request) else {
+            throw TestFailure("expected .invalidParameterValue for tool_choice naming unknown tool")
+        }
+        try assertTrue(detail.contains("foo"))
+        try assertTrue(detail.contains("bar"))
+    }
+
+    test("validator accepts named tool_choice when name matches a tool (#480)") {
+        let request = try decode(
+            ChatCompletionRequest.self,
+            from: #"{"model":"\#(M)","messages":[{"role":"user","content":"hi"}],"tools":[{"type":"function","function":{"name":"lookup","description":"d"}}],"tool_choice":{"type":"function","function":{"name":"lookup"}}}"#
+        )
+        try assertNil(ChatRequestValidator.validate(request))
+    }
+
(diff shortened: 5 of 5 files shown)


Arthur-Ficial (commented):

Scope is honest and the scope-split in the description is the right instinct — #480 is genuinely an L-effort issue and slicing it is better than a mega-PR. Three things to fix before this merges, one of them load-bearing.

### 1. `Fixes #480` will close an issue that is 1/5 done

The description itself lists four remaining steps (response-boundary gating, `parallel_tool_calls: false` enforcement, typed `tool_choice_not_satisfied` errors, `supported_parameters`). But `Fixes #480` in the body is a GitHub closing keyword — merging this auto-closes #480 and the remaining 80% of the work silently disappears from the open list. **Change it to `Refs #480`** (or `Part of #480`), and let Franz close the issue when step 5 lands.

### 2. The backwards-compatible initializer overload is unnecessary and slightly risky

Unlike #488 — where `max_completion_tokens: Int?` has *no* default and the overload is genuinely required — here `parallel_tool_calls: Bool? = nil` is already defaulted in the designated initializer. Source compatibility for every existing caller is therefore preserved without the overload. What the extra initializer actually buys you is two candidates that differ only by a defaulted parameter, which is exactly the shape that produces confusing overload-resolution behaviour later. Recommend deleting it and keeping the single designated init.

### 3. Merge-order hazard with #488 — this one will break the build

#488 adds a backwards-compatible initializer overload with the **identical 20-parameter legacy signature** to the one this PR adds. They conflict in git, and if a conflict resolution keeps both, the result is an *invalid redeclaration* and will not compile. Whichever of the two merges second must fold its new field into the other's designated initializer and drop its own overload — then rebase and re-run CI. The current green check on this branch does not survive #488 landing first.

### What is good

- **Not adding `parallel_tool_calls` to `supported_parameters` is the right call**, and I want to say so explicitly since it looks like an omission. The field is decoded but not yet enforced; advertising it now would be exactly the kind of "fake capability" CLAUDE.md's *Honest about limitations* rules out. Correct to defer to step 5.
- The named-`tool_choice` validation is a real fix — a request naming a function absent from `tools` previously reached the model with an instruction to call something that was never defined.
- Test coverage is proportionate: 3 decoding tests, 4 validation tests, and the existing #238 `tool_choice` format tests are untouched and still valid (they send `tool_choice` without `tools`, which the new guard deliberately skips).

### One behaviour to document in code, not just in the PR body

The named-choice check is skipped when `tools` is nil or empty, because MCP-injected tools resolve after validation. That is a sound decision, but it means the impossible-name case this PR fixes is *still* reachable whenever the server runs with `--mcp` and the client names a tool that no MCP server provides. The rationale currently lives only in this PR description, where the next reader of `ChatRequestValidator.swift` will not find it. A two-line comment at the guard would carry it, and the residual MCP gap belongs in step 2's scope.

### On verification

CI is green on `macos-26` — `build-and-test`, `apfelcore-public-api`, `changelog-gate` — so the new unit tests genuinely run and pass; the "cannot run FoundationModels" caveat applies to the authoring runner, not to CI. What remains unproven is live end-to-end behaviour with the real on-device model. `Test plan` checkboxes are unticked and need @franzenzenhofer locally.

Cheers, Arthur

Arthur-Ficial:
Good catches, all three. Here's where things stand after this push (58ba7a9):

**1. `Fixes #480` → `Refs #480`.** Done. PR body updated. The issue stays open until step 5 lands.

**2. The backwards-compatible init overload.** I'd like to delete it too, but `swift package diagnose-api-breaking-changes` won't let me. My first commit (c9168e3) had only the designated init with `parallel_tool_calls: Bool? = nil` defaulted - `apfelcore-public-api` failed with:

> API breakage: constructor `ChatCompletionRequest.init(model:messages:...:tool_choice:response_format:...)` has been removed

The tool compares symbol signatures, not source compatibility - adding a parameter between `tool_choice` and `response_format` changes the label list and the old symbol disappears from the API dump. The overload restores it. I agree it's not ideal for overload resolution, but it's the price of the CI gate. If Franz would rather take the API break (it's additive from a source perspective), the check would need an allowlist entry for this symbol - that's a CI change I shouldn't make from a routine.

**3. Merge-order with #488.** Noted in the PR body now. Whichever merges second folds its new field into the other's designated init and drops its duplicate overload. Straightforward git conflict, not a logic conflict.

**4. Code comment for the MCP-skip rationale.** Added in 58ba7a9 - two lines at the guard explaining why `tools` nil/empty is skipped and that the residual MCP gap belongs in step 2.

Cheers, Arthur

---

Arthur-Ficial:
Release triage 2026-09-19: leaving this one open and not including it in today's release. Reasons: it is step 1 of the 5-step #480 plan, so `parallel_tool_calls` is decoded but still not enforced (a half-implementation by the project's own rule), and the finish line (response-boundary gating, typed errors, `supported_parameters`) needs a design pass rather than a bugfix pass. It also needs a rebase now that #488 landed on `ChatCompletionRequest.init`. The named-`tool_choice` validation half is sound and can be split out if someone wants it in the next patch. Cheers, Arthur

Arthur-Ficial (commented):
Daily triage 2026-09-20 — first-pass re-review. **The merge-order conflict this PR predicted has now happened.**

The PR body's own merge-order note said: "#488 adds a backwards-compatible init overload with the same 20-parameter legacy signature. Whichever merges second must fold its new field into the other's designated init."

#488 merged 2026-09-19 and shipped in **v1.11.0**. This PR is now `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `main`. The three green checks (`build-and-test`, `changelog-gate`, `apfelcore-public-api`) are from a stale base and no longer tell us anything about merging today.

**Golden-goal read: aligned.** Both halves serve the OpenAI-compatible server contract — decoding a field the API advertises, and rejecting a named `tool_choice` that can never be satisfied. Honest 400s over silent wrong behaviour is exactly the stated principle. The MCP-skip carve-out (skip the check when `tools` is nil/empty) is the right call and is documented at the guard.

**Process notes, unchanged and still correct:**
- Stays a draft. It is step 1 of 5 on #480 — steps 2-5 (response-boundary enforcement, `parallel_tool_calls: false` enforcement, typed errors, `supported_parameters`) are still open.
- The body correctly says `Refs #480`, not `Fixes` — so merging won't prematurely close a 5-step issue. Good.
- "Functional correctness not verified" is still true and still the honest framing: this triage machine cannot run FoundationModels.

**What's needed before this can be looked at again:** a rebase onto `main` @ v1.11.0, folding the `parallel_tool_calls` field into the designated init that #488 left behind, then a green CI run on the new base. I don't push branches on unattended runs, so that's a Franz call or a follow-up session.

No merge recommendation from me — @franzenzenhofer decides, and it needs a local `make test` on Apple Intelligence hardware first.

Cheers, Arthur

Arthur-Ficial:
Daily triage 2026-09-24 — status update to my 09-20 re-review: **the merge-order conflict is resolved.**

About 40 s after that review, `fix(core): preserve pre-#480 ChatCompletionRequest init for API stability` landed. It adds the v1.11.0-signature overload that delegates with `parallel_tool_calls: nil`. The current head is `MERGEABLE`, and all three checks pass, including `apfelcore-public-api`. That was the gate my review was worried about.

Nothing blocks this on the triage side now. It's still a draft by design, so it needs @franzenzenhofer's review plus the local test run the body asks for. This is step 1 of the #480 scope-split. The actual `parallel_tool_calls: false` enforcement at the response boundary is still open in #480.

Cheers, Arthur

Arthur-Ficial (commented):
## Re-review after synchronize (2026-09-24) — `COMMENTED`, not a merge signal

The PR has been rebased onto main (post-v1.11.0) and expanded from step 1 of 5 to covering the full #480 scope: request-level validation **and** response-boundary enforcement, `parallel_tool_calls` decoding + enforcement, typed violation errors, and `supported_parameters` advertising. The merge conflict the Sept 20 triage flagged is resolved and the v1.11.0 backward-compatible overload is folded in correctly. Scope is now aligned with the updated title and body.

### What is good

- **`ToolPolicy` is clean architecture.** A pure `ApfelCore` struct with no external dependencies, `Sendable + Equatable`, and a `Result`-returning `resolve` that separates request-scope validation (400s) from response-boundary enforcement (`evaluate` → `Outcome`). The two-phase design means the same policy object is resolved once and used on all three server paths (non-streaming, streaming, MCP auto-execute) plus Responses — no duplicated logic.

- **Streaming forced-choice buffering is the right call.** `holdAllContent = jsonMode || policy.forcesToolCall` ensures no partial content leaks to the client before the verdict is known. Without this, a forced-choice violation would arrive after content deltas were already delivered, breaking the HTTP 200 envelope.

- **MCP path correctly uses `enforceNames: false`.** Unknown tool names on the MCP path are left to `MCPManager`, which feeds a tool-not-found error back to the model for recovery (#241 behavior preserved). Mode and cap enforcement still apply — the test `"enforceNames false defers unknown names to the executor but still applies mode and cap"` verifies this.

- **The `executeMCPToolCallsForServer` refactoring is sound.** Taking pre-validated `firstRound: [ParsedToolCall]` instead of raw content cleanly separates detection + policy from execution. Subsequent MCP re-prompt rounds correctly remain unconstrained — `tool_choice` is a first-generation contract, not a loop invariant.

- **Backward-compatible init chain is correct.** Three overloads (designated with both `max_completion_tokens` + `parallel_tool_calls`, v1.11.0 compat with `max_completion_tokens` only, pre-v1.11.0 with neither) preserve source compatibility. `swift package diagnose-api-breaking-changes main` confirms no breaks.

- **Responses streaming path correctly skips ToolPolicy.** The validator rejects `tools with stream: true` on `/v1/responses`, so no tool-call detection or policy evaluation is needed there.

- **Test coverage is proportionate and TDD-shaped.** 22 unit tests for the pure policy logic cover every resolve case, every evaluate outcome, the `enforceNames` toggle, and `maxCalls` capping. 4 model-free server validation tests (400s, `/v1/models`). 4 model-backed client tests including `tool_choice: "none"` on every seed and the `required` contract. 3 MCP tests including the MCP-dispatch-disabled check via server log. The `conftest.py` seed rotation now handles `tool_choice_not_satisfied` 500s like guardrail refusals — correct, since both are model trajectory outcomes.

- **Documentation updated in all three places** — CHANGELOG, `tool-calling-guide.md` (new enforcement table), `openai-api-compatibility.md` (new row). The model-quirks table entry for hallucinated tool names is updated to reflect the new enforcement behavior.

### Findings

| Priority | Area | Finding |
|----------|------|---------|
| P2 | `ToolPolicy` | Disabled mode carries a meaningless `maxCalls` when `tool_choice: "none"` + `parallel_tool_calls: false` — `evaluate` returns `.content` before reaching the cap check. Not a bug, just a dead value. |
| P2 | Duplicate check | `ChatRequestValidator` (line 263) and `ToolPolicy.resolve` (line 128) both validate named `tool_choice` against tools when client tools are present. Validator runs first, so ToolPolicy's check is unreachable in that case. Defense in depth, not wrong — but the error messages differ slightly ("tools array" vs "tools in scope"). |
| P2 | MCP re-prompt rounds | `parallel_tool_calls: false` caps the first MCP round via `firstRound` but subsequent rounds in the `while true` loop use `detectAndExecuteMCPTools` without capping. If the model generates multiple tool calls in a follow-up round, all are executed. Arguable whether `parallel_tool_calls` should govern internal re-prompt rounds (the client only sees the final text), but worth documenting the decision. |

All three are P2 — follow-up PR acceptable, none blocking.

### CI status at review time

- `changelog-gate`: ✅ passed
- `build-and-test`: ⏳ in progress
- `apfelcore-public-api`: ⏳ in progress

CI started ~20 minutes ago on the new base. Cannot verify results yet. The PR author reports all 1186 unit tests passing locally with `diagnose-api-breaking-changes` clean.

### What remains

- **CI green on this base** — the Sept 20 review correctly noted the stale checks are meaningless; these new ones are the real signal.
- **Local `make test` on Apple Intelligence hardware** — functional correctness for the model-backed tests (tool_choice none actually suppressing MCP dispatch, required actually producing a 500 on plain text, etc.) is not verifiable from CI.
- **Merge decision is Franz's** — this is a substantial behavior change to the OpenAI-compatible server contract. The error semantics (typed 500 instead of silent 200) are correct per OpenAI's spec, but clients that relied on the old permissive behavior will see new failures.

No P0 or P1 findings. Waiting on CI before any merge recommendation.

---

Issue #196, opened 2026-06-08
Watch: re-check Core AI beta docs at macOS 27 GM / release


Arthur-Ficial:
Part of epic #189.

The Core AI docs at https://developer.apple.com/documentation/coreai/ are **Beta** as of 2026-06-09.
Symbol availability metadata is inconsistent (several pages omit macOS/watchOS in their availability
blocks while the framework root lists them) and the surface may shift before GM.

### Tasks
- [ ] At macOS 27 GM (and again at public release), re-fetch the Core AI docs and diff against the
      snapshot in [docs/coreai-impact.md](docs/coreai-impact.md).
- [ ] Re-confirm the FoundationModels-vs-Core-AI boundary still holds (no LLM/chat/tool APIs migrated
      into Core AI that would change the conclusion).
- [ ] Update the knowledge page + this epic if the conclusion changes; otherwise close as "no change".
- [ ] Re-verify the WWDC 2026 FoundationModels claims (context window, base model) are now documented
      fact, and unblock #_(context)_ / #_(base model)_ accordingly.

### Note
This is the heartbeat task that keeps the epic honest as the beta matures. Low effort, recurring.

Arthur-Ficial:
Daily-vet 2026-06-14: clean, well-scoped heartbeat task under epic #189. Correctly labeled `documentation`.

This fits the "honest about limitations" principle — the whole point is to keep `docs/coreai-impact.md` and the OS 27 conclusions accurate as the Core AI beta matures, rather than baking in speculation. The task list (re-fetch + diff docs, re-confirm the FoundationModels-vs-Core-AI boundary, unblock #192/#193) is the right shape.

Not actionable yet: Core AI docs are still Beta and the unblock steps depend on macOS 27 GM/release hardware, which I don't have from a cloud runner. This stays a recurring watch item — re-run the diff at GM and again at public release, then close as "no change" or file deltas.

Cheers, Arthur

Arthur-Ficial:
## Watch heartbeat 2026-09-09 — gate still closed, macOS 27 has not shipped

This is a watch ticket and it had gone **three months** without a check (last note 2026-06-14), so here is a dated one.

**Checked from a Mac, not from press coverage:**

```
$ sw_vers
ProductVersion:  26.5.2   BuildVersion: 25F84

$ softwareupdate --list
* Label: macOS Tahoe 26.6.2-25G83   Version: 26.6.2   Recommended: YES
```

[...]

PR #493, opened 2026-09-16
fix(reporting): distinguish measured vs assumed context window (#491)

### LINKED ISSUE #491: macOS 27: --model-info and /health report the hardcoded 4096 floor as a measured context window
On macOS 27, `--model-info` and `/health` present the hardcoded 4096 floor as if it were a measured context window.

Found while reading #490 / #196 on a macOS 27 GA machine.

## Environment

| | |
|---|---|
| macOS | **27.0 (26A428)** — the GA build |
| Chip | Apple M4 Pro |
| Toolchain | Swift 6.4 (`swiftlang-6.4.0.34.1`), SDK 27.0, target `arm64-apple-macosx27.0.0` |
| apfel | `main` @ f8942de, `swift build` clean (exit 0) |
| Apple Intelligence | **not enabled on this machine** — see "What I did not verify" |

## What I measured

`SystemLanguageModel.contextSize` read straight from the SDK, no apfel in the path:

```swift
import FoundationModels
let m = SystemLanguageModel.default
print(m.isAvailable)    // false
print(m.availability)   // unavailable(...UnavailableReason.modelNotReady)
print(m.contextSize)    // 0
```

Raw `contextSize` is **`0`**. apfel then reports:

```
$ apfel --model-info
├ available:  no (model not ready - still downloading?)
├ context:    4096 tokens
```

That `4096` is not a reading. It is the floor in `Sources/TokenCounter.swift:72`, printed by
`Sources/CLI.swift:613` with no indication that it is a fallback rather than a measurement.

## Why this matters on 27 specifically

The floor is documented as *"the known minimum for any Apple Intelligence model"*
(`Sources/TokenCounter.swift:58-60`). That was true on macOS 26. Per #192 and
`docs/coreai-impact.md`, the macOS 27 window is **8192** — so on 27 the assumed value is
**half** the real one.

This is invisible on macOS 26, where the floor happens to equal the true window. It only
becomes observable on 27, which is why I think it survived the #192 and #330 sweeps.

`/health` has the same problem, and it is the one that actually bites: `Sources/Server.swift:90`
feeds the identical value into the wire response, and `/v1/models` does the same at
`Sources/Server.swift:117`. An OpenAI-compatible client that autodiscovers the window gets
`4096` and sizes `max_tokens` to half the machine's capacity.

This is not only the unavailable case. Per the project's own comment at `Sources/Server.swift:65-68`,
on macOS 27 the SDK returns 0 *"during initialization (~80s cold start)"* — so there is a real
window in which the model is warming and `/health` serves an assumed number as fact.

## Why I am not proposing to remove the floor

The floor is correct and should stay. It prevents the #192 deadlock where `inputBudget` goes
negative, generation is rejected, and the model never warms up. **The defect is in the reporting
layer, not the budget math** — a fallback is being rendered indistinguishably from a measurement.

This is what CLAUDE.md's *"Never bake a single context-size literal into code or user-facing prose"*
and *"Honest about limitations"* are pointing at.

## Suggested fix (additive, non-breaking)

`context_window` is pinned as `{"type": "integer"}` by `Tests/integration/openapi_spec_test.py:218`
and asserted `> 0` by `test_health_context_window_positive` (#192). So it must **not** become
nullable. Instead:

- Carry "was this measured or assumed?" alongside the number.
- `--model-info`: mark the assumed case, e.g. `context: 4096 tokens (assumed - model not ready)`.
- `/health` and `/v1/models`: add a sibling boolean, e.g. `context_window_measured`. Additive, so
  the existing spec tests and clients keep working.

The natural home for the decision is a small pure type in `ApfelCore` — the `apfel-tests` target
depends on `ApfelCore`/`ApfelCLI` only, so anything inside `TokenCounter` (FoundationModels target)
cannot be unit-tested.

Happy to send a PR with tests first if the approach looks right.

## What I did NOT verify

- **I did not measure 8192 myself.** Apple Intelligence is not enabled on this machine, so the
  `UAF_FM_GenerativeModels` asset is absent. The 8192 figure here is the project's own #192
  finding, not an independent confirmation. I am enabling Apple Intelligence and will follow up
  with the real GA reading, including whether the cold-start-0 behaviour still occurs on 26A428.
- I have not run the model-marked integration tests for the same reason.

Arthur-Ficial:

Fixes #491.

## Root cause

`TokenCounter.contextSize` applies a high-water mark and a 4096-token floor to `SystemLanguageModel.contextSize` to prevent the #192 deadlock where `inputBudget` goes negative. The floor is correct. But the three reporting surfaces - `--model-info` (CLI.swift:613), `/health` (Server.swift:90), and `/v1/models` (Server.swift:117) - all present the floor indistinguishably from a measured SDK reading. On macOS 26 this is invisible (floor equals the real window), but on macOS 27 where the window is 8192 the assumed value is half the actual capacity. An OpenAI-compatible client autodiscovering via `/v1/models` would size `max_tokens` to half the machine's real budget.

## The fix

Adds a pure `ContextWindow` struct in ApfelCore carrying `size: Int` and `measured: Bool`. `TokenCounter` gains a `contextWindow` property that returns this type (the existing `contextSize: Int` delegates to `contextWindow.size` so all budget-math call sites are unchanged). The three reporting surfaces now use `contextWindow`:

- `--model-info`: shows `4096 tokens (assumed - model not ready)` when the value is the floor
- `/health`: adds `context_window_measured: false` alongside the existing `context_window` integer
- `/v1/models`: same - adds `context_window_measured` to the model object

The wire change is additive: `context_window` stays a required integer (existing spec tests and clients keep working), the boolean is a new sibling field.

## Test coverage

- Added `Tests/apfelTests/ContextWindowTests.swift` - four unit tests for init, equality, and Sendable conformance.
- Added `ContextWindow` to `ApfelCorePublicAPIUsageTests.swift` compile-time lockdown.
- Added `test_health_context_window_measured_present` and `test_models_context_window_measured_present` integration tests in `openapi_spec_test.py`.
- Updated `HEALTH_SCHEMA` to require `context_window_measured` as a boolean.
- Existing `test_health_context_window_positive` and `test_models_context_window_positive` still cover the > 0 invariant.

- All existing `contextSize` call sites (CLI --count-tokens, Benchmark, inputBudget) go through the delegating `contextSize: Int` property - no signature change, no breakage.
- `ModelsListResponse.ModelObject` gains the new field as an `Encodable` property - additive, no wire breakage.
- Swift 6 concurrency: `ContextWindow` is `Sendable` (all-`let` struct of `Int` + `Bool`). No new actor boundaries or `@unchecked Sendable`.
- Diff limited to `Sources/Core/ContextWindow.swift`, `Sources/TokenCounter.swift`, `Sources/CLI.swift`, `Sources/Server.swift`, `Sources/Models.swift`, tests, and `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

Nothing - straightforward bug report from a contributor with a clear reproducer. No embedded code to copy, no shell commands to run, no claims of prior authorization.

---

https://claude.ai/code/session_01Pu3wFt5HhGtqrgVF6juJD2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pu3wFt5HhGtqrgVF6juJD2)_


wyleung (commented):
> **Conflict of interest, up front:** I filed #491 and opened #494, which fixes the same bug. Both
> PRs add `Sources/Core/ContextWindow.swift` defining a `ContextWindow` type, so they cannot both
> merge. Please weigh everything below with that in mind. I have tried to make every claim
> reproducible rather than asking anyone to take my word for it, and I am happy to close #494 —
> see "Suggested path forward".

Tested on macOS 27.0 (`26A428`), M4 Pro, Swift 6.4 / SDK 27 — the hardware CI cannot provide.

## What works

The design is right, and it is the right shape for this codebase:

- **The floor is kept.** The PR correctly identifies that the #192 deadlock guard must survive and
  only the *reporting* changes. That was the easy thing to get wrong.
- **`contextSize: Int` delegating to `contextWindow.size`** is a genuinely nice touch — every budget
  call site (`inputBudget`, `--count-tokens`, `Benchmark`) is untouched, and I verified that holds:
  274 model-free integration tests pass.
- **The wire change is additive.** `context_window` stays a required integer, so #192's guarantee and
  existing clients are unaffected.
- **`HEALTH_SCHEMA` was updated to *require* the new field**, not merely allow it. Stricter than it
  had to be, and correct.
- **The public-API lockdown entry** in `ApfelCorePublicAPIUsageTests.swift` is exactly the sort of
  thing that is easy to skip.
- CHANGELOG entry present; `changelog-gate` green.

**And the fix genuinely works.** On this macOS 27 machine:

```
$ curl -s localhost:11434/health | jq '{context_window, context_window_measured}'
{ "context_window": 4096, "context_window_measured": false }

$ apfel --model-info
├ context:    4096 tokens (assumed - model not ready)
```

## Findings

| # | Sev | Area | Summary |
|---|-----|------|---------|
| 1 | P1 | Tests | The unit tests cannot fail. Reintroducing #491 exactly leaves all 1124 green. |
| 2 | P1 | Tests | Both new integration tests assert presence and type, never the value their docstrings promise. |
| 3 | P2 | UX | `(assumed - model not ready)` can print directly under `available: yes`. |
| 4 | P2 | Comment | The file header claims the distinction is unit-testable; it is not. |

---

### 1. (P1) The unit tests cannot fail

I injected one mutation into `Sources/TokenCounter.swift` — the assumed branch now claims it was
measured, which *is* bug #491, restored verbatim:

```diff
-        return ContextWindow(size: 4096, measured: false)
+        return ContextWindow(size: 4096, measured: true)
```

Result:

```
$ swift run apfel-tests
✅ All 1124 tests passed
```

The bug this PR exists to fix is back, and the suite is silent.

**Root cause.** `ContextWindow` is a data carrier: two `let`s, a memberwise init, synthesized
`Equatable`. So the four tests in `ContextWindowTests.swift` check that the Swift compiler stores
what you pass it, compares two `Int`/`Bool` pairs, and conforms a struct of value types to
`Sendable`. None of that is apfel's logic.

The actual decision — *raw 0 ⇒ not measured, raw > 0 ⇒ measured, high-water survives a regression to
0* — stayed in `TokenCounter.contextWindow` (`Sources/TokenCounter.swift:64-78`), which lives in the
FoundationModels target. `apfel-tests` depends on `ApfelCore` and `ApfelCLI` only
(`Package.swift:61-65`), so nothing in there is reachable from a unit test. The type moved to
`ApfelCore`; the logic did not.

**Concrete fix** — move the decision, not just the shape, and have `TokenCounter` hold one:

```swift
public struct ContextWindowTracker: Sendable {
    public static let assumedTokens = 4096
    private var highWater = 0
    public init() {}

    public mutating func observe(_ rawReading: Int) -> ContextWindow {
        if rawReading > highWater { highWater = rawReading }
        guard highWater > 0 else { return ContextWindow(size: Self.assumedTokens, measured: false) }
        return ContextWindow(size: highWater, measured: true)
    }
}
```

```swift
// TokenCounter
private var _windowTracker = ContextWindowTracker()
var contextWindow: ContextWindow { _windowTracker.observe(model.contextSize) }
```

The test that actually pins the bug is the one asserting that a **measured 4096 and an assumed 4096
are the same number but not the same claim** — that is the whole defect in one assertion, and no
test in this PR makes it.

As a control, I ran the equivalent mutation against #494, where the logic does live in `ApfelCore`:
3 tests fail. I mention that only to show the difference is the placement of the logic, not a
difference of opinion about test style.

### 2. (P1) The integration tests assert presence, not behaviour

`test_health_context_window_measured_present` docstring says:

> context_window is the assumed 4096 floor and context_window_measured **must be false**. When the
> SDK has reported a real value, it **must be true**.

The test asserts neither. It checks `in data` and `isinstance(..., bool)`. Against the mutant above,
the server served:

```json
{ "context_window": 4096, "context_window_measured": true }
```

and both new tests passed:

```
$ pytest Tests/integration/openapi_spec_test.py -k context_window
4 passed, 30 deselected
```

**Concrete fix.** There is a model-free invariant available that is deterministic on CI: if the model
is not available, the SDK cannot have reported a positive size, so the value must be assumed.

```python
if not data["model_available"]:
    assert data["context_window_measured"] is False, (
        "model unavailable means contextSize was never read, so the window is the assumed floor (#491)"
    )
```

That fails against the mutant and passes against this PR's real code. It also runs on GitHub CI,
where the model is never available.

### 3. (P2) `(assumed - model not ready)` can contradict the line above it

`Sources/CLI.swift:606-608` ties the hedge to "model not ready", but the two facts are independent:
`isAvailable` is `model.isAvailable` (`Sources/TokenCounter.swift:87-88`), while the flag comes from
`contextSize`. Per this repo's own note at `Sources/Server.swift:65-68`, on macOS 27 the SDK returns
0 *"during initialization (~80s cold start)"* — so there is a window where the model is up and the
size has not landed, and `--model-info` would print:

```
├ available:  yes
├ context:    4096 tokens (assumed - model not ready)
```

Wording that does not assert a cause avoids it, e.g. `(assumed - the model has not reported a size yet)`.

**Caveat:** I could not reproduce this. Apple Intelligence is not enabled on this machine, so I never
reach the cold-start window. This is from reading the code, not from observing it.

### 4. (P2) The header comment overclaims

`Sources/Core/ContextWindow.swift:4-5`:

> Pure ApfelCore type so the measured-vs-assumed distinction is unit-testable without FoundationModels

The *carrier* is in ApfelCore; the distinction is not, as finding 1 shows. Worth correcting whichever
way the code goes, so the next reader does not trust it.

## What I verified

- `swift build` and `swift build -c release` — clean, no new warnings beyond the three pre-existing SDK 27 ones (`Session.swift:52` deprecation, two `.reasoning(_)` exhaustiveness warnings; see #196 comment).
- `swift run apfel-tests` — 1124 passed.
- `pytest Tests/integration/ -m "not model"` — **274 passed, 6 skipped, 1 failed**.
- The one failure is `test_chat.py::test_chat_non_tty_rejected` (expects exit 2, gets exit 5). **Pre-existing and unrelated to this PR** — it reproduces on `main` and on an unrelated branch. The availability guard fires before the TTY check, so it fails on any machine without Apple Intelligence. Separately, it is not marked `@pytest.mark.model` despite depending on the model being available.
- Real behaviour on macOS 27 hardware is correct (`measured: false` plus the hedged CLI line, both shown above).
- The `#192` regression tests still pass, so the floor still holds.
- Budget call sites unaffected by the `contextSize` delegation.

## What I did NOT verify

- The 239 `@pytest.mark.model` tests, and the `measured: true` path — Apple Intelligence is not
  enabled here, so `contextSize` never leaves 0. I have only ever observed the assumed branch. The
  measured branch is verified by code reading alone, in this PR and in mine.
- Finding 3, as noted.

## Suggested path forward

Ranked, and I genuinely do not mind which:

1. **Keep this PR, port the tests.** It is yours, it is already CI-green, and the production diff is
   sound. Move the decision into a tracker in `ApfelCore` (finding 1), tighten the two integration
   assertions (finding 2), and I will close #494. Happy to push those onto this branch if that is
   easier.
2. **Take #494 instead**, which already has the logic and tests in that shape, and close this.
3. **Neither** — merge this as-is if the test gap is acceptable to you. The production behaviour is
   correct today; the exposure is that a future refactor of `TokenCounter` can silently undo it with
   a green suite.

Option 1 is the one I would pick.

_Reviewed and tested with [Claude Code](https://claude.com/claude-code)._

Arthur-Ficial:
Triage follow-up on this draft (my own bug-solver output).

@wyleung's review above is correct, and I've now checked it rather than taking it on trust. `ContextWindowTests.swift` here only constructs the struct and compares fields. The measured/assumed decision stays in `TokenCounter` (FoundationModels target), which `apfel-tests` can't reach, so a regression to #491 would pass the unit suite. The two integration tests also assert only presence and type. On #494, where the logic lives in `ApfelCore.ContextWindowTracker`, the equivalent mutation fails 3 tests on this machine (macOS 26.5.2). I also saw `context_window_measured: true` live with Apple Intelligence enabled.

**Recommendation for @franzenzenhofer:** take #494 and close this draft. The two conflict (both add `Sources/Core/ContextWindow.swift`). #494 is only missing the `ApfelCorePublicAPIUsageTests` lockdown entry, which this draft did include. I'm leaving the close to Franz rather than closing it myself.

Cheers, Arthur

---

Arthur-Ficial:
Closing in favour of #494 by @wyleung, which is now merged: same wire shape, but the measured/assumed decision lives in ApfelCore.ContextWindowTracker where the unit suite can pin it (a mutation there fails 3 tests; the version here could not catch a regression). Credit to @wyleung for the fix and the review that made the call clear. Cheers, Arthur

PR #494, opened 2026-09-16
fix(context): stop reporting the assumed window floor as a measurement (#491)

### LINKED ISSUE #491: macOS 27: --model-info and /health report the hardcoded 4096 floor as a measured context window

wyleung:
Fixes #491.

## Summary

`TokenCounter.contextSize` falls back to a 4096-token floor when
`SystemLanguageModel.contextSize` reports 0. **That floor is correct and this PR keeps it** —
without it `inputBudget` goes negative, generation is rejected, and the model never warms up (#192).

What was wrong is that the fallback was rendered indistinguishably from a real reading. On macOS 26
that is invisible, because the floor equals the true window. On macOS 27 the true window is 8192, so
`--model-info`, `/health` and `/v1/models` each reported half the machine's real capacity as fact —
and `/health` is what an OpenAI-compatible client reads to size `max_tokens`.

Before, on this macOS 27 machine:

```
$ curl -s localhost:11434/v1/models | jq '.data[0].context_window'
4096
```

After:

```
$ curl -s localhost:11434/health | jq '{context_window, context_window_measured}'
{ "context_window": 4096, "context_window_measured": false }

$ apfel --model-info
├ context:    4096 tokens (assumed - the model has not reported a size yet)
```

## Design

The measured/assumed decision lives in a new pure `ApfelCore.ContextWindowTracker`. That is
deliberate: the `apfel-tests` target depends on `ApfelCore` and `ApfelCLI` only, so nothing inside
`TokenCounter` (FoundationModels target) can be unit-tested. Moving the decision into `ApfelCore`
is what makes it coverable. The high-water-mark behaviour from #192 moves with it unchanged and is
now under test for the first time.

`context_window` stays a positive integer on both endpoints, so the #192 guarantee, the pinned
`HEALTH_SCHEMA` (`{"type": "integer"}`) and existing clients are unaffected. The new
`context_window_measured` field is additive, which `additionalProperties: True` already allows.

New public `ApfelCore` API (`ContextWindow`, `ContextWindowTracker`) is additive, so
`diagnose-api-breaking-changes` stays clean.

## Test plan

TDD throughout, red before green:

1. `ContextWindowTests.swift` written first — failed with `cannot find 'ContextWindowTracker' in scope`.
2. Implemented `ApfelCore/ContextWindow.swift` → green.
3. `displayText` tests added — failed with `value of type 'ContextWindow' has no member 'displayText'`.
4. Implemented → green.
5. The two integration tests added — failed with `AssertionError: /v1/models must expose context_window_measured (#491)`.
6. Implemented the server fields → green.

The sharpest test is the one that pins the actual defect: a *measured* 4096 and an *assumed* 4096
are the same number but not the same claim, and only the flag separates them.

Results on macOS 27.0 (26A428), M4 Pro, Swift 6.4 / SDK 27:

- `swift run apfel-tests` — **1129 passed** (1119 + 10 new)
- `pytest Tests/integration/ -m "not model"` — **265 passed, 12 skipped**
- `-k context_window` (the #192 regression tests plus the two new ones) — **4 passed**
- `swift build` and `swift build -c release` — no new warnings

## What I verified

- The #192 tests (`test_health_context_window_positive`, `test_models_context_window_positive`)
  still pass, so the floor still holds.
- `HEALTH_SCHEMA` still validates; `context_window` is still an integer.
- `CHANGELOG.md` has an `## [Unreleased]` bullet, per the `changelog-gate` job (#369).
- No touches to `.version`, `BuildInfo.swift`, the README badge, or any release infrastructure.

## What I did NOT verify

- **I have not measured the real 8192 window myself.** Apple Intelligence is not enabled on this
  machine, so the `UAF_FM_GenerativeModels` asset is absent and `contextSize` reads 0. The 8192
  figure comes from the project's own #192 finding. This PR does not depend on the exact number —
  it only makes the assumed case distinguishable from a measured one.
- The 239 `@pytest.mark.model` tests, for the same reason.
- One pre-existing failure, `test_chat.py::test_chat_non_tty_rejected`, expects exit 2 but gets
  exit 5 (model unavailable) on this machine. It is unrelated to this diff — the availability guard
  fires before the TTY check — but worth flagging: it is *not* marked `@pytest.mark.model` despite
  its assertion depending on the model being available. I am happy to open a separate marker-
  discipline ticket if you would like.

I am enabling Apple Intelligence and will follow up on #491 with the real GA reading, including
whether the cold-start-0 behaviour still occurs on 26A428.


Arthur-Ficial (commented):
First-pass review (automated daily triage). **COMMENT only - not an approval, and I do not merge.**

### Golden-goal fit

Squarely on-goal: this is the server surface (`/health`, `/v1/models`) plus the "Honest about limitations" principle. Presenting a guessed window as a measurement is exactly the kind of quiet dishonesty that principle rules out. The fix is additive on the wire, and the #192 floor survives.

### Verified locally - complementing your macOS 27 run

You could only observe the *assumed* branch. This Mac has Apple Intelligence enabled, so I covered the other half:

| | |
|---|---|
| macOS | 26.5.2, Apple Silicon, Apple Intelligence **on** |
| `swift build` | clean |
| `swift run apfel-tests` | **1129 passed** |
| `apfel --model-info` | `context:    4096 tokens` (no hedge, correct) |
| `/health` | `{"context_window":4096,"context_window_measured":true,"model_available":true}` |
| `/v1/models` | `{"context_window":4096,"context_window_measured":true}` |

Mutation check: I flipped the assumed branch in `ContextWindowTracker.observe` to `isMeasured: true`, which puts #491 back, and got **3 failed, 1126 passed**. So the tests pin the defect, and the "measured 4096 vs assumed 4096" test is the right sharp edge. Between your run and mine, both branches have now been observed on real hardware.

### Findings (none blocking)

| # | Sev | Summary |
|---|-----|---------|
| 1 | P2 | `ContextWindow` and `ContextWindowTracker` are new `public` ApfelCore API but aren't added to `Tests/apfelTests/ApfelCorePublicAPIUsageTests.swift`, the compile-time lockdown file that exercises every public symbol. #493 did add its entry. Worth porting, including `displayText`, `assumedTokens` and `observe(_:)`. |
| 2 | P2 | The two new integration tests assert presence and type only. That's the same gap you rightly flagged on #493 (finding 2), so it applies here too. Your proposed model-free invariant (`if not data["model_available"]: assert data["context_window_measured"] is False`) is deterministic on CI and belongs in this PR. |
| 3 | P2 | Optional: `HEALTH_SCHEMA` could list `context_window_measured: boolean`. #493 required it. Not needed, since `additionalProperties: True`. |

Release note for the reviewer: STABILITY.md (line 42) lets bug-fix-motivated additive ApfelCore API ship in a PATCH (precedent: `TokenCountFallback`, v1.7.1).

### Process

Returning contributor (#38, #54, #56 merged). One commit on a feature branch, no merge commits. `CHANGELOG.md [Unreleased]` entry present, all checks green, no version/badge/`BuildInfo.swift` edits.

### Relationship to #493

#493 is my own bug-solver draft for the same issue. Both add `Sources/Core/ContextWindow.swift`, so only one can land. Having checked both, **I recommend this PR** over mine: the decision logic sits where the unit tests can reach it, and #493's tests can't catch a regression. Franz makes the call.

Cheers, Arthur

---

PR #490, opened 2026-09-15
docs(coreai-impact): drop stale 'all Beta' after macOS 27 GA (#196)

Arthur-Ficial:
**Draft PR - needs @franzenzenhofer review before merging.**

Refs #196.

## Summary

Core AI shipped GA with macOS 27.0 on 2026-09-14. The beta flag on every Core AI platform symbol now reads `beta: false`, confirmed against the GA docs by the #196 watch heartbeats (09-11 RC, 09-15 GA). This removes the three stale beta references from `docs/coreai-impact.md`:

- Header: "(Core AI, beta)" -> "(Core AI)", plus a dated re-verification note
- Availability line: "27.0+, all Beta" -> "27.0+", updated to note GA shipment date
- Sources section: "live beta JSON docs" -> reflects GA status

## Root cause

The page was written 2026-06-09 when Core AI was still in beta. macOS 27 GA landed 2026-09-14 and every Core AI symbol now reads `beta: false`. The doc's "all Beta" language became stale at that point.

## Test plan

- Docs-only change - no build or test run needed.
- [x] Verified all three edits remove only the stale beta language.
- [x] Confirmed no links were broken.
- [x] No `Sources/**` files touched, so no CHANGELOG entry required.

## What I verified (static only)

- Diff limited to `docs/coreai-impact.md` - one file, three targeted edits.
- No touches to release infrastructure, guardrails, CI, or dependencies.
- No code changes, no new dependencies, no forbidden files modified.

## What I did NOT verify

- N/A for a docs-only change. No functional verification needed.

## Anything that seemed suspicious

Nothing - straightforward documentation freshness update driven by the #196 watch heartbeat series.

## Note on issue closure

Changed from `Fixes #196` to `Refs #196` per triage feedback: SDK 27 GA runtime checks found new drift (`Transcript.Entry.reasoning(_)`, `GenerationOptions(sampling:)` deprecation) that #196 should continue tracking until those are ticketed. The doc edits here are still correct; #196 stays open for the remaining items.

---

cc @franzenzenhofer - draft stays draft until you review. The #196 watch heartbeats (09-11 through 09-15) have the full evidence trail.

_Generated by the apfel bug-solver routine._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WqKSZ1eMAt1VqVYMU61KGR

Issue #330, opened 2026-07-02
ApfelError contextOverflow openAIMessage hardcodes 4096 while the window size is dynamic everywhere else


Arthur-Ficial:
`ApfelError.openAIMessage` for `.contextOverflow` (`Sources/Core/ApfelError.swift:141`) hardcodes the window size:

```swift
return "Input exceeds the 4096-token context window. Shorten the conversation history."
```

Every runtime path elsewhere reads the dynamic `TokenCounter.shared.contextSize`, and `/health`, `--count-tokens`, etc. report the live value. This one string is truthful today (26.5.2 still reports 4096 - runtime-verified) but becomes a lie the day the OS reports a different window (docs/coreai-impact.md:155-171 predicts 8192 on macOS 27). `Tests/apfelTests/ApfelErrorMessageTests.swift:36` pins the wrong-by-then string.

**Fix:** drop the number from the message ("Input exceeds the model's context window...") or thread the actual size into the error at the call sites. Update the pinned test. Related to the OS-27 tracking issues #189/#192 but this is a concrete code fix, not a docs sweep.

Found in the v1.7.1 bug sweep.

Arthur-Ficial:
Fix drafted - see #340. Drops the hardcoded "4096-token" from the `contextOverflow` message so it stays truthful regardless of OS-reported window size. Needs @franzenzenhofer to review and run the tests locally before merging.

Cheers, Arthur

cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

Arthur-Ficial:
Fix drafted - see #342. Drops the hardcoded "4096-token" from the `contextOverflow` message so it stays honest when the OS reports a different window size. Pinned test updated to match.

Needs @franzenzenhofer to review and run the tests locally before merging.

Cheers, Arthur

cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

Arthur-Ficial:
Fix drafted - see #356. Dropped the hardcoded "4096-token" from the `contextOverflow` message, now reads "Input exceeds the model's context window." Chose the number-free approach over threading the size as an associated value since `.contextOverflow` is a simple enum case in the public `ApfelCore` API. Updated the pinned test to match.

Needs @franzenzenhofer to review and run `swift run apfel-tests` locally before merging.

Cheers, Arthur
cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

[...]

PR #495, opened 2026-09-16
docs(integrations): make the copy-paste token limits macOS-27-correct

wyleung:
Docs-only. Found while checking #490 / #196 on a macOS 27 GA machine.

## Summary

Three integration pages ship copy-paste configs pinned to macOS 26 token limits, and one of them
over-commits the window regardless of OS.

### `docs/vscode-copilot.md` — the actual bug

```json
"maxInputTokens": 4096,
"maxOutputTokens": 4096
```

followed by *"The `4096` token limits match the macOS 26 on-device window."*

They do not match it. The window is **input and output combined** — `TokenCounter.inputBudget` is
`contextSize - reservedForOutput` (`Sources/TokenCounter.swift:76`), and #192 describes it as
*"input + output combined"*. So this asks for 8192 tokens of a 4096-token window, and it is wrong on
macOS 26 and macOS 27 alike.

Split into a macOS 26 block (3584 + 512) and a macOS 27 block (7680 + 512), each summing to its
window and leaving apfel's default 512-token reserve intact. Two blocks rather than one annotated
block, per CLAUDE.md's *"One code block, one purpose"* rule — a reader must copy one or the other,
so an inline comment would not be safe to paste.

### `docs/local-setup-with-vs-code.md` and `docs/integrations.md` — stale, not wrong

Both already carried the correct macOS 27 figure **in prose**, but left the snippet at `4096`. A
macOS 27 user who copies the block — which is the point of the block — silently runs at half the
window, and the prose reads as background rather than as an instruction. Both now state plainly what
to change and what it costs if you do not.

## What I verified

- `Sources/TokenCounter.swift:76` — `inputBudget(reservedForOutput: Int = 512) -> contextSize - reservedForOutput`, confirming the combined window and the 512 default.
- `apfel --help` — `--context-output-reserve <n>` exists, so the closing note points at a real flag.
- `test_examples_doc.py`, `test_changelog_gate.py`, `test_marker_discipline.py` — 9 passed.
- `scripts/check-changelog.sh` — *"no production source changes - CHANGELOG entry not required"*.
- No `Sources/**` touched; no links added or changed.

## What I did NOT verify

- I did not exercise the Copilot or Continue UIs against these configs. The numbers follow from
  apfel's own budget arithmetic, not from a round-trip through either extension. If Copilot treats
  `maxInputTokens`/`maxOutputTokens` as independent hints rather than a budget that must sum to the
  window, the split is still safe but more conservative than strictly necessary.
- Apple Intelligence is not enabled on this machine, so I could not confirm the 8192 figure myself —
  it is the project's own #192 finding. Context on #196.


Arthur-Ficial (commented):
First-pass review (automated daily triage). **COMMENT only - not an approval, and I do not merge.**

### Golden-goal fit

Docs-only, off the README, and it serves the "OpenAI-compatible server as a drop-in" goal: these are the copy-paste configs people actually paste. The Copilot correction is right. The window covers input and output combined (`TokenCounter.inputBudget` = `contextSize - reservedForOutput`), so `4096 + 4096` over-asks on every OS. For Zed, `max_tokens` is the model's total context and `max_output_tokens` is separate, so the rewording is accurate. Same for Continue's `contextLength`.

### Finding

**P1 - the closing note in `docs/vscode-copilot.md` recommends a command that exits 2:**

```
$ apfel --serve --port 11498 --context-output-reserve 1024
Error: --serve does not accept --context-output-reserve - it would be ignored in this mode
exit=2
```

`--context-output-reserve` exists (so `apfel --help` lists it), but `CLIArguments.swift:297-312` rejects it in `--serve` mode on purpose. In the server, the reserve is per request: `x_context_output_reserve` in the request body, falling back to `BodyLimits.defaultOutputReserveTokens` (`Handlers.swift:136`). As far as I know, Copilot's BYOK config can't send that extension field. So the practical guidance is simply: *keep `maxOutputTokens` at 512 (apfel's server default), or if you raise it, lower `maxInputTokens` so the two still sum to the window.* Suggest dropping the `apfel --serve --context-output-reserve <n>` sentence.

### Optional, out of scope

The Zed snippet in `docs/integrations.md` still has `max_output_tokens: 1024` against the server's default 512 reserve. Not introduced here, and Zed trims differently, so I'm only noting it. No need to touch it in this PR.

### Process

Docs-only, so no CHANGELOG owed (`changelog-gate` green). One commit on a feature branch. It pairs the macOS 26 and 27 values with an `apfel --model-info` pointer, consistent with how the repo already handles the context-size-literal rule (#330).

Franz decides merge; I've only reviewed.

Cheers, Arthur

---

Issue #370, opened 2026-07-09
feat: support -f for --chat as well


mblarsen:
Support -f arguments for --chat as well

Arthur-Ficial:
Hey @mblarsen, thanks, genuinely good idea.

I had a look at the code path and you're right - `-f` content gets parsed and merged into the prompt string in `main.swift`, but the `.chat` case at line 345 calls `chat()` without passing that merged prompt through. So `apfel -f code.swift --chat` silently discards the file content. Not great.

The natural fix would be to inject the file content as the first user turn (or as additional context alongside the system prompt) when chat mode starts, so you'd land in the interactive session with the file already in context. Something like `apfel -f code.swift --chat` giving you a conversation where the model has already seen the file.

This fits the CLI chat side of apfel nicely. Labelling as `enhancement` - Franz decides priority from here.

Cheers, Arthur

cc @franzenzenhofer

---
_Generated by [Claude Code](https://claude.ai/code)_

Arthur-Ficial:
Done and shipped in **v1.8.2** — https://github.com/Arthur-Ficial/apfel/releases/tag/v1.8.2

Thanks @mblarsen — credited you as co-author.

`apfel -f file.swift --chat` now folds the extracted file (and any positional/piped content) into the chat session so the model has it in context on the very first turn, with a `context:` notice on startup. Before, the `.chat` dispatch read/OCR'd the file and then silently discarded it.

While fixing it we audited the whole silent-drop class: `--serve`, `--benchmark`, `--model-info`, and `--update` also parsed and then ignored a prompt / `-f` / `-s` / generation flags. Those now reject with a clear usage error (exit 2) instead of pretending to accept them. `--serve` still honors its real flags.

Update: `brew upgrade apfel`

Cheers, Arthur

PR #497, opened 2026-09-19
fix(cli): downgrade env-sourced APFEL_* defaults to warnings in input-ignoring modes (#496)

### LINKED ISSUE #496: fix(cli): exported APFEL_* prompt defaults make --serve exit 2
## Problem

Prompt-mode defaults exported in the shell make `apfel --serve` refuse to start with exit code 2, blaming a flag the user never passed:

```
$ APFEL_CONTEXT_OUTPUT_RESERVE=1024 apfel --serve
error: --serve does not accept --context-output-reserve - it would be ignored in this mode
$ echo $?
2
```

The same happens for `APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, `APFEL_MAX_TOKENS`, `APFEL_CONTEXT_STRATEGY` and `APFEL_CONTEXT_MAX_TURNS`, and for the other input-ignoring modes (`--benchmark`, `--model-info`, `--update`).

## Root cause

`CLIArguments.parse` applies `APFEL_*` env defaults into the same struct fields as the flags. The #370 guard in `validate(context:)` (`Sources/CLI/CLIArguments.swift`, "inputIgnoringModes") then rejects any non-nil input field in those modes without knowing whether it came from an explicit flag or from the environment. Rejecting a flag that would be silently ignored is right; rejecting an exported default is not - the documented purpose of those variables is to be a standing default for prompt modes, so a user who exports one for daily CLI use cannot start the server at all.

## Expected

- Explicit flags keep the hard error (unchanged).
- Env-derived defaults in an input-ignoring mode are dropped with the existing non-fatal warning pattern (#254), e.g. `warning: ignoring APFEL_CONTEXT_OUTPUT_RESERVE in --serve mode`, and the command runs.

Found on 2026-09-19 while checking the closing note of #495.

Arthur-Ficial:

Fixes #496.

### Root cause

`CLIArguments.parse()` applies `APFEL_*` env vars into the same struct fields as CLI flags. The #370 guard in `validate()` then rejected any non-nil input field in `--serve`/`--benchmark`/`--model-info`/`--update` without knowing whether it came from an explicit flag or the environment. A user who exported `APFEL_CONTEXT_OUTPUT_RESERVE` (or any prompt-mode default) for daily CLI use could not start the server at all - it exited 2 blaming a flag the user never passed.

### The fix

A new `envSourcedFields: Set<String>` on `CLIArguments` tracks which optional fields were populated from env vars during parsing. When a CLI flag explicitly overrides an env default, the tag is removed. In `validate()`, the input-ignoring-modes guard now iterates all offending fields: env-sourced ones produce a non-fatal warning (e.g. `warning: ignoring APFEL_CONTEXT_OUTPUT_RESERVE in --serve mode`) and are nil'd out; explicit flags keep the existing hard error unchanged. `validate()` becomes `mutating` to support this.

### Test coverage

- Added 18 new tests in `Tests/apfelTests/CLIValidateTests.swift`:
  - 6 `validate()` unit tests for each env-sourced field/mode combination (systemPrompt/serve, temperature/serve, contextOutputReserve/serve, maxTokens/benchmark, contextStrategy/model-info, contextMaxTurns/update)
  - 1 test that explicit flags still throw even when other fields are env-sourced
  - 1 test that multiple env-sourced warnings are collected in one pass
  - 1 test that an explicit flag error fires even when env-sourced warnings precede it in the checks
  - 7 `parse()` end-to-end tests for each env var with `--serve`
  - 1 `parse()` end-to-end test for `--benchmark` with multiple env defaults
  - 1 test that `--serve --temperature 0.5` (explicit flag) still rejects
  - 1 test that `--serve --temperature 0.5` with `APFEL_TEMPERATURE` env (explicit overrides env) still rejects
  - 1 regression test that env-sourced fields work normally in prompt modes
- Existing `CLIValidateTests` updated for `mutating validate()` (trivial `var` binding changes)

## Test plan
- [ ] `swift run apfel-tests`
- [ ] `APFEL_CONTEXT_OUTPUT_RESERVE=1024 apfel --serve` starts with a warning (not exit 2)
- [ ] `apfel --serve --temperature 0.5` still exits 2

- Source and test code read in full; fix is consistent with existing patterns.
- Swift 6 concurrency: `Set<String>` is `Sendable`; `CLIArguments` remains `Sendable, Equatable`.
- No data races introduced - `validate()` mutates only via the owning `var`.
- Diff limited to `Sources/CLI/CLIArguments.swift`, `Tests/apfelTests/CLIValidateTests.swift`, `CHANGELOG.md` - no touches to release infrastructure, guardrails, CI, or dependencies.

## What I did NOT verify

Nothing - straightforward bug from the project owner. Clean report, no injection.

---

https://claude.ai/code/session_01CXVL3ef1btX9wyse3FM3jp

---
_Generated by [Claude Code](https://claude.ai/code/session_01CXVL3ef1btX9wyse3FM3jp)_


Arthur-Ficial:
Thanks - this landed at the same time as my own fix for #496 (#498), which I had already driven red-to-green on this Mac. Going with #498 because it keeps the change inside parse(): validate() stays non-mutating and CLIArguments gains no new public field, so the ApfelCLI surface is untouched, and it adds a model-free e2e test. Closing this one as superseded. Cheers, Arthur

Issue #254, opened 2026-07-01
fix(env): invalid APFEL_* environment values are silently ignored while equivalent flags hard-error


Arthur-Ficial:
## Summary
Invalid env values are silently dropped to defaults, while the equivalent flags exit 2. `APFEL_PORT=99999 apfel --serve` binds 11434 without a word; `APFEL_CONTEXT_STRATEGY=newest_first` (typo, underscore) silently uses the default. Verified: `APFEL_TEMPERATURE=abc APFEL_PORT=99999 APFEL_CONTEXT_STRATEGY=bogus apfel --count-tokens hi` → exit 0, no warning; `--port 99999` → exit 2.

## Root cause
`Sources/CLI/CLIArguments.swift:203-213` — `APFEL_PORT` out-of-range → silent 11434; `APFEL_TEMPERATURE` non-numeric/negative → silent nil; `APFEL_CONTEXT_STRATEGY` unknown → silent nil; same for `APFEL_MAX_TOKENS`, `APFEL_MCP_TIMEOUT`, `APFEL_CONTEXT_MAX_TURNS`, `APFEL_CONTEXT_OUTPUT_RESERVE`.

## Suggested fix
Have `parse()` collect env-validation warnings (e.g. a `[String]` of "ignoring APFEL_PORT=99999 (not in 1-65535)") on the returned struct; main.swift prints them to stderr unless `--quiet`. Keeps `parse()` pure and testable.

## Severity
Medium (individually low, systematic).

---
_Filed from a full read-only audit of apfel v1.6.1 (server, CLI/chat, MCP/tool-calling, security, concurrency, tests/CI/docs). Findings verified against committed code; high-severity crash/DoS items were reproduced against the release binary. One issue per finding._

Arthur-Ficial:
Fixed in 51b501d - invalid APFEL_* values (port range, non-numeric temperature, unknown strategy, etc.) now produce stderr warnings (suppressed by --quiet) instead of silent fallback; parse() stays pure and collects warnings on the result. Ships in the next release.

PR #498, opened 2026-09-19
fix(cli): drop exported APFEL_* prompt defaults in input-ignoring modes instead of exiting 2 (#496)

### LINKED ISSUE #496: fix(cli): exported APFEL_* prompt defaults make --serve exit 2

Arthur-Ficial:
Fixes #496.

## Root cause

`CLIArguments.parse` applies `APFEL_*` env defaults into the same fields as the flags. The #370 guard in `validate(context:)` then rejected any non-nil prompt/tuning field in `--serve`, `--benchmark`, `--model-info` and `--update`, so a user with `APFEL_CONTEXT_OUTPUT_RESERVE` (or any of the five other prompt defaults) exported for daily CLI use could not start the server: exit 2, blaming a flag they never passed.

## Fix

`parse()` records which of those flags were explicit; in an input-ignoring mode the env-derived values are cleared with a #254-style warning (`warning: ignoring APFEL_CONTEXT_OUTPUT_RESERVE in --serve mode (it only applies to prompt modes)`). `validate()` is unchanged, so explicit flags keep the hard error and the existing #370 tests still hold.

## Tests (TDD, red first)

- 6 unit tests in `CLIArgumentsTests.swift`: each env default dropped with a warning in `--serve`; same for `--benchmark`/`--model-info`/`--update`; explicit `--temperature` and `-s` still hard-error when the env var is also set; env defaults still apply and never warn in prompt modes. 3 failed before the fix with the exact #370 message, all green after.
- 1 model-free e2e test in `cli_e2e_test.py`: `--model-info` with two env defaults exits 0 and prints both warnings.
- CHANGELOG `[Unreleased]` bullet added.


Arthur-Ficial (commented):
## Review — PR #498: fix(cli): drop exported APFEL_* prompt defaults in input-ignoring modes (#496)

### Summary

Clean, well-scoped fix for a real usability bug. No P0, P1, or P2 findings.

| Severity | Area | Count |
|----------|------|-------|
| P0 | — | 0 |
| P1 | — | 0 |
| P2 | — | 0 |

### What the fix does

`parse()` now tracks which prompt-related flags were explicitly passed on the command line via `explicitFlags`. After the parse loop completes but before `validate()` runs, env-derived prompt defaults (`APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, `APFEL_MAX_TOKENS`, `APFEL_CONTEXT_STRATEGY`, `APFEL_CONTEXT_MAX_TURNS`, `APFEL_CONTEXT_OUTPUT_RESERVE`) are cleared with a warning in input-ignoring modes — but only when no explicit flag for that same setting was passed. `validate()` is unchanged, so an explicit `--temperature 0.5` on `--serve` still hard-errors exactly as #370 intended.

### What I verified

- **Root cause matches #496:** `parse()` populates struct fields identically from env vars and flags; `validate()` can't tell them apart. The fix correctly splits the two paths.
- **`envPromptDefaults` table is complete:** All six env-var defaults parsed in lines 376–437 are represented. `APFEL_TOP_P` and `APFEL_SEED` don't exist as env vars (confirmed via grep), so `topP`/`seed` are flag-only and correctly left to `validate()`.
- **`--system-file` edge case:** Listed in `envPromptDefaults` flags for `APFEL_SYSTEM_PROMPT`, so `explicitFlags` correctly records it. If a user passes `--serve --system-file foo.txt` with `APFEL_SYSTEM_PROMPT` exported, the cleanup skips the entry and `validate()` catches the non-nil `systemPrompt` with the #370 error. ✓
- **Empty env var handling:** `env[entry.env].map({ !$0.isEmpty }) == true` mirrors the `envValue()` helper at line 372 — an empty string is treated as absence. ✓
- **`inputIgnoringModes` promotion:** The `Set<Mode>` moved from a local in `validate()` to a `static let` shared by both `validate()` and the cleanup block. The set contents are identical. ✓
- **`validate()` unchanged:** The #370 guard (lines 307–328) is byte-identical except the `Self.inputIgnoringModes` reference. All five existing #370 integration tests (`test_serve_rejects_positional_prompt`, `test_serve_rejects_system_prompt`, `test_serve_rejects_generation_flag`, `test_benchmark_rejects_positional_prompt`, `test_model_info_rejects_tuning_flag`) pass explicit CLI flags, not env vars, so they're unaffected. ✓
- **Warning format:** Follows the #254 pattern (`ignoring APFEL_... in --serve mode`), consistent with existing warnings. ✓
- **No concurrency concerns:** `envPromptDefaults` and `inputIgnoringModes` are immutable static lets on a `Sendable` struct. `explicitFlags` is a local variable. ✓
- **CHANGELOG `[Unreleased]` entry present.** ✓
- **No security surface changes.** ✓

### Test coverage

All six unit tests exercise distinct scenarios with proper TDD style:

1. Single env default → dropped + warning in `--serve` (the core bug)
2. All six env defaults → all dropped, six warnings in `--serve`
3. Coverage across `--benchmark`, `--model-info`, `--update` (all input-ignoring modes)
4. Explicit `--temperature` + env `APFEL_TEMPERATURE` → still hard-errors (regression guard for #370)
5. Explicit `-s` + env `APFEL_SYSTEM_PROMPT` → still hard-errors (short-flag variant)
6. Prompt modes (`["hi"]`, `["--chat"]`, `["--stream", "hi"]`) → env defaults apply, no warnings (no regression)

The e2e integration test uses `--model-info` (model-free, exits at validation), so it runs in CI. ✓

### CI

- `changelog-gate`: ✅ passed
- `build-and-test`: in progress
- `apfelcore-public-api`: in progress

No issues found. Well done — the fix is minimal, the test matrix is thorough, and the existing #370 guard is preserved.

---

Arthur-Ficial (commented):
## Re-review after synchronize (rebase onto updated main)

Diff is substantively identical to the previously reviewed commit (`a0f13979`). Re-verified on `7399595`:

| Severity | Area | Count |
|----------|------|-------|
| P0 | — | 0 |
| P1 | — | 0 |
| P2 | — | 0 |

### What I re-checked

- **`envPromptDefaults` table** still covers exactly the six env vars parsed at lines 376–436 (`APFEL_SYSTEM_PROMPT`, `APFEL_TEMPERATURE`, `APFEL_MAX_TOKENS`, `APFEL_CONTEXT_STRATEGY`, `APFEL_CONTEXT_MAX_TURNS`, `APFEL_CONTEXT_OUTPUT_RESERVE`). Server/operational vars (`APFEL_PORT`, `APFEL_HOST`, `APFEL_TOKEN`, `APFEL_MCP`, `APFEL_MCP_TIMEOUT`, `APFEL_MCP_TOKEN`, `APFEL_DEBUG`) correctly excluded. ✓
- **`explicitFlags` tracking** (line 504) runs at the top of each loop iteration before the `switch`, capturing the flag token; value arguments like `"0.5"` after `--temperature` don't match any entry, so they're correctly ignored. ✓
- **Cleanup ordering** — env-derived values nil'd out after the parse loop (line 893) but before `validate()` (line 916), so the #370 guard sees a clean struct for env-only cases. ✓
- **Empty-string treatment** — `env[entry.env].map({ !$0.isEmpty }) == true` matches the `envValue()` helper pattern. ✓
- **`--system-file` edge case** — listed in `envPromptDefaults` flags for `APFEL_SYSTEM_PROMPT`, so `explicitFlags` records it and the entry is not cleaned, letting `validate()` reject the explicit flag. ✓
- **`inputIgnoringModes` promotion** — set contents identical to the former local, now shared between `validate()` (line 313) and the cleanup block (line 893). ✓
- **Concurrency** — immutable `static let`s on a `Sendable` struct; `explicitFlags` is local to `parse()`. ✓
- **Test coverage** — 6 unit tests + 1 model-free e2e test, all with tightened assertions. ✓
- **CHANGELOG `[Unreleased]` entry** present. ✓

### CI status

- `changelog-gate`: ✅ passed
- `build-and-test`: 🔄 in progress
- `apfelcore-public-api`: 🔄 in progress

No new findings. Fix remains clean.

---

Arthur-Ficial (commented):
## Re-review after synchronize (`6971d32`)

Diff is substantively identical to the two prior reviews. No new findings.

| Severity | Count |
|----------|-------|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |

### Verified on this head

- **`envPromptDefaults` table** covers all six prompt-related env vars parsed at lines 376–437; server/operational vars (`APFEL_PORT`, `APFEL_HOST`, `APFEL_TOKEN`, `APFEL_MCP`, `APFEL_MCP_TIMEOUT`, `APFEL_MCP_TOKEN`, `APFEL_DEBUG`) correctly excluded. ✓
- **`explicitFlags` tracking** captures the flag token before the `switch`; value args (e.g. `"0.5"` after `--temperature`) don't match any entry. ✓
- **Cleanup ordering** — env-derived values nil'd after parse loop (line 888) but before `validate()` (line 916), so #370 guard sees a clean struct for env-only cases. ✓
- **Empty-string treatment** — `env[entry.env].map({ !$0.isEmpty }) == true` mirrors the `envValue()` helper. ✓
- **`--system-file` edge case** — listed in `envPromptDefaults` flags, so explicit `--system-file` is tracked and cleanup is skipped, letting `validate()` catch it. ✓
- **New e2e test** correctly lacks `@pytest.mark.model` (model-free, consistent with adjacent `test_model_info_rejects_tuning_flag`). ✓
- **Concurrency** — immutable `static let`s on `Sendable` struct; `explicitFlags` local. ✓
- **CHANGELOG `[Unreleased]`** entry present. ✓

### CI

All three checks (`changelog-gate`, `build-and-test`, `apfelcore-public-api`) in progress on this head.

---

Issue #189, opened 2026-06-08
Epic: FoundationModels OS 27 update (new model + LanguageModel/CoreAILanguageModel bridge) - impact on apfel


Arthur-Ficial:
## Epic: FoundationModels OS 27 update - the real on-device-AI change for apfel

WWDC 2026 (keynote 2026-06-08) surfaced two things. Only one actually matters for apfel, and this
epic now leads with it.

### The real story (this epic)

**FoundationModels - the framework apfel is built on - gets a substantial update in OS 27.** Confirmed
by Apple's official [Foundation Models updates](https://developer.apple.com/documentation/updates/foundationmodels)
page (June 2026 entries), not press speculation:

- **New on-device model** (reportedly Gemini-distilled). Apple says verbatim: *"the model changes when
  a person updates to iOS 27 ... macOS 27 ... test your prompts with the new model."* apfel inherits any
  shift in tool-call formatting, refusals, tokenization, token counts.
- **`LanguageModel` protocol + open-source `CoreAILanguageModel` / `MLXLanguageModel`** - an official
  bridge to drive any model (server, on-device, even non-Apple weights via Core AI) through the
  existing FoundationModels session API. This is what makes a bring-your-own-model path tractable.
- **`GenerationOptions.ToolCallingMode`** and **improved error types** (`LanguageModelError`,
  `SystemLanguageModel.Error`, `LanguageModelSession.Error`) - direct adoption candidates for apfel's
  tool layer and `ApfelError.classify`.
- **On-device context window still reads 4096.** The larger window Apple advertises is the **cloud**
  `PrivateCloudComputeLanguageModel`, which apfel does not use. So apfel's 4096 docs likely stay correct.
- Also: `DynamicProfile` agentic API, image analysis (`OCRTool`/`BarcodeReaderTool`), Python SDK -
  out of scope for apfel's golden goal.

### The non-event (context, not the headline)

"**Core AI**" is the rebrand / successor of **Core ML** - a low-level tensor inference runtime
(`AIModel` / `NDArray` / `InferenceFunction`). It does **not** replace FoundationModels and needs **no
apfel-core change on its own**. It matters here only as the runtime behind the new `CoreAILanguageModel`
bridge above. Full clarification + sources:
[docs/coreai-impact.md](https://github.com/Arthur-Ficial/apfel/blob/main/docs/coreai-impact.md).

### Scope / decision

- apfel stays on FoundationModels. No migration.
[...]

PR #499, opened 2026-09-21
fix(build): pin --build-system native for CLT-only environments (#194)

### LINKED ISSUE #194: macOS 27 SDK build + full test-suite qualification
Part of epic #189.

apfel pins `platforms: [.macOS(.v26)]` in `Package.swift`. macOS 27 (the OS that ships Core AI and the
OS-27 FoundationModels) needs an explicit build + runtime qualification pass.

### Tasks
- [ ] Build apfel against the macOS 27 / Xcode 27 SDK - clean, no new warnings, Swift 6 strict
      concurrency still satisfied.
- [ ] Confirm FoundationModels availability gates and `@available` annotations still hold on 27.
- [ ] Confirm `SystemLanguageModel.tokenCount(for:)` (SDK 26.4+ API) and `GenerationOptions` are
      unchanged or adapt if not.
- [ ] Run `make preflight` (599 unit + integration suites, 0 skipped) green on macOS 27 hardware.
- [ ] Decide whether to keep `.macOS(.v26)` (back-compat) or bump - and update the "macOS 26 Tahoe
      required" gotcha messaging if the floor moves.
- [ ] Verify Homebrew `depends_on :macos` / `:tahoe` symbol still resolves correctly on 27.

### Gating
Blocked on macOS 27 SDK + hardware availability.

Arthur-Ficial:

Fixes #194 (the CLT-only build break identified by @halostatue in the issue discussion).

## Summary

SwiftPM 6.4 changed the default `--build-system` from `native` to `swiftbuild` ([swiftlang/swift-package-manager#9661](https://github.com/swiftlang/swift-package-manager/pull/9661)). Under Command Line Tools only, the `swiftbuild` backend dies at init with "Unknown error parsing property list." The Makefile had 5 bare `swift build`/`swift run` calls, all of which broke.

## Root cause

SwiftPM 6.4's default build-system flip. CLT-only environments lack the Xcode project support that `swiftbuild` needs. CI never caught it because `ci.yml` explicitly selects the newest Xcode before building.

## The fix

- All `swift build -c release` calls in the Makefile (build, release-patch, release-minor, release-major) and the `swift run apfel-tests` call in the test target now include `--build-system native`.
- `scripts/check-no-xcode.sh` updated to match.
- `docs/install.md` updated: the "build from source" section clarifies that `make install` handles the workaround automatically, and a new troubleshooting section covers the "Unknown error parsing property list" error for users running `swift build` directly.
- `--build-system native` is deprecated and temporary - the upstream fix is tracked at [swiftlang/swift-package-manager#10561](https://github.com/swiftlang/swift-package-manager/issues/10561).

## Test coverage

- Added `test_swift_build_calls_use_native_build_system` and `test_swift_run_calls_use_native_build_system` to `Tests/integration/test_build_info.py` to lock the invariant - any future `swift build` or `swift run` added to the Makefile without `--build-system native` fails CI.
- Existing `test_build_target_does_not_depend_on_generate_build_info` and `test_release_targets_still_depend_on_generate_build_info` still pass.

- Diff limited to `Makefile`, `docs/install.md`, `scripts/check-no-xcode.sh`, `CHANGELOG.md`, `Tests/integration/test_build_info.py` - no touches to release infrastructure, guardrails, CI, or dependencies.
- Policy test passes against the fixed Makefile, fails against the original (TDD red-green confirmed).
- Swift 6 concurrency: no Swift source changed.
- No forbidden files touched.

## What I did NOT verify

- **Functional correctness. This needs `make test` on a Mac with Apple Intelligence.** I cannot run FoundationModels code or `swift build` in this cloud environment. If the fix does not actually resolve the reported behaviour, that is on me to refine - ping me in a review comment.

Nothing - straightforward build-system compatibility fix. The bug was reported by a MacPorts maintainer with a clear upstream reference.

## Notes for @franzenzenhofer

- `scripts/publish-release.sh` and `scripts/release-preflight.sh` also have bare `swift run apfel-tests` calls that likely need the same fix, but those are release infrastructure I cannot touch from a routine. Worth checking.
- The `--build-system native` flag is deprecated upstream. When the SwiftPM fix lands, the flag should be removed (the policy test will remind you it exists).
- The triage comment noted MacPorts is not listed in `docs/install.md` - separate concern, not included here.

## Test plan
- [ ] `swift build -c release --build-system native`
- [ ] `swift run --build-system native apfel-tests`
- [ ] Verify under CLT-only (`sudo xcode-select -s /Library/Developer/CommandLineTools`) that `make install` succeeds
- [ ] `make test` full suite

---

https://claude.ai/code/session_01J4P51GHjgNTEkzEJmp43Jd

---
_Generated by [Claude Code](https://claude.ai/code/session_01J4P51GHjgNTEkzEJmp43Jd)_


Arthur-Ficial (commented):
First-pass review (daily triage). Scored against the Golden Goal: apfel is a UNIX tool + OpenAI-compatible server, and "build from source with Command Line Tools" is a documented install path — so a CLT-only build break is squarely in scope and worth fixing. The change is build-only, touches no Swift source, and keeps the two core products untouched. Alignment: good.

**I ran the verification this PR says it could not.** This Mac is CLT-only (`xcode-select -p` → `/Library/Developer/CommandLineTools`), macOS 26.5.2, Swift 6.3.3. On the `pr499` branch:

| Check | Result |
|---|---|
| `swift build -c release --build-system native` | Build complete (98.88s) |
| `swift run --build-system native apfel-tests` | All 1157 tests passed |
| `pytest Tests/integration/test_build_info.py` | 4 passed |
| Same pytest against **main's** Makefile | 2 failed, as designed — red-green confirmed |

The red check is the useful one: the two new policy tests genuinely fail on the unfixed Makefile (`Makefile:206: swift run without --build-system native`), so the invariant is really locked, not just asserted.

**One honest limit on the above:** our toolchain here is SwiftPM 6.3.3, where `native` is *already* the default. So what I proved is that the pin is **harmless and backward-compatible on the documented Swift 6.3+ minimum** — not that it cures the 6.4 break. That half still rests on @halostatue's report and the upstream issue, which I consider solid. I did check that `--build-system` parses on 6.3.3 (it does, `default: native`), which was the main regression risk of pinning a flag: users on the documented minimum are unaffected.

**One blocking-ish issue: `Fixes #194` is the wrong keyword and the wrong issue.**

#194 is *"macOS 27 SDK build + full test-suite qualification"* — a six-task child of epic #189, still blocked on macOS 27 hardware. This PR fixes none of those six tasks. The CLT break only *lives* in #194 because @halostatue reported it in that thread. Merging as-is auto-closes an unfinished, hardware-blocked epic child. This is the same trap flagged on #489 (`Fixes #480` on a 5-step ticket).

Suggested: open a dedicated issue for the CLT-only build break and have this PR say `Fixes #<new>` / `Refs #194`. I have not opened that issue myself — issue creation is outside what I do unattended. @franzenzenhofer's call.

**Second gap — the fix is incomplete and the new tests can't catch it.** Both policy tests scan `Makefile` only, but two bare `swift run` calls survive:

- `scripts/publish-release.sh:51` → `swift run apfel-tests`
- `scripts/release-preflight.sh:81` → `if swift run apfel-tests 2>&1; then`

So cutting a release from a CLT-only machine still breaks on SwiftPM 6.4+, and nothing fails CI to tell you. The PR body flags these as untouchable release infrastructure, which I respect — but worth either fixing in a follow-up or widening the test's glob to `scripts/*.sh` so the gap is at least visible. `check-no-xcode.sh` was correctly updated.

**Checked and clear:** `Makefile:261` runs `swift package clean`, which the tests skip — correct, `swift package clean` takes no `--build-system` flag, so that's a non-issue, not a coverage hole.

Net: mechanically sound, well-tested, verified green on real CLT-only hardware. Fix the `Fixes #194` keyword before this merges. Staying a draft is right — the 6.4-specific behaviour still wants one run on a SwiftPM 6.4 toolchain, which we don't have here.

Cheers, Arthur

cc @franzenzenhofer

Arthur-Ficial:
Nice catch on the `Fixes #194` keyword - you're right, merging this would auto-close the macOS 27 qualification tracker. I've updated the PR body to `Refs #194` and added a note that Franz should open a dedicated CLT-break issue for this PR to close instead.

On the release scripts gap: agreed, `publish-release.sh:51` and `release-preflight.sh:81` are real. I've added a "Known gaps" section to the PR body calling them out explicitly. Widening the test to scan `scripts/*.sh` would make CI red on files this routine can't touch, so I've left the test scoped to `Makefile` and flagged the rest for Franz.

The backward-compatibility verification on SwiftPM 6.3.3 is exactly the data I couldn't produce - good to have that confirmed.

Cheers, Arthur

---

Arthur-Ficial:
Follow-up from the same triage — **the blast radius is wider than this repo.** Worth knowing before you decide how to land this.

I swept every Swift repo in the family for unpinned `swift build` / `swift run` / `swift test` calls in their Makefiles. **Twelve repos are affected, and this PR fixes one of them:**

| Repo | Unpinned calls | Documents build-from-source? |
|---|---|---|
| apfel | 5 | yes (this PR) |
| apfelfw | 11 (via `SWIFT_BUILD_FLAGS` / `SWIFT_TEST_FLAGS`) | yes |
| apfel-gui | 4 | yes |
| translate | 4 | yes |
| ohr | 3 | yes |
| auge | 3 | yes |
| bgbgone | 3 | yes |
| apfel-quick | 3 | yes |
| apfel-run | 2 | yes |
| apfel-chat | 2 | yes |
| apfel-tag | 2 | yes |
| apfelpad | 2 | yes |

(`kern` and `apfel-server-kit` have no Makefile — clean.)

Every one of those READMEs tells users to `make install` or `swift build` from source, so on SwiftPM 6.4+ under CLT-only they all fail the same way, with the same unhelpful `Unknown error parsing property list`. The ecosystem CLAUDE.md lists *"Same build system … Makefile with auto-bump, `make install`"* as a shared principle — which is exactly why the bug is shared too.

Two notes that make the sweep cheaper:

- **`swift test` takes the flag as well** (verified on 6.3.3: `--build-system <build-system> (default: native)`). Six of those repos use `swift test` rather than `swift run`, so a copy-paste of this PR's policy test would miss them — the test here only matches `swift build` / `swift run`.
- **apfelfw is nearly free**: all 11 of its calls already route through `SWIFT_BUILD_FLAGS` / `SWIFT_TEST_FLAGS`, so adding `--build-system native` to those two variables fixes the whole repo in two lines.

Not proposing anything here — this PR is correctly scoped to apfel and shouldn't grow. Flagging it so the other eleven don't get forgotten once this merges and the symptom disappears from view. @franzenzenhofer, if you want a tracking issue on apfel-ecosystem for the family-wide sweep, say the word and I'll open one — I don't create issues on an unattended run.

Cheers, Arthur

Arthur-Ficial:
Review on real CLT-only hardware (macOS 26.5, Swift 6.3.3, `xcode-select -p` = /Library/Developer/CommandLineTools), branch head 1bca5a2.

The fix is right and narrowly scoped: build-only, no Swift source touched, and the `--build-system` flag parses on the documented Swift 6.3 minimum (`default: native`), so pinning is a no-op there and a fix on 6.4+.

What I added on the branch before merging:
- The two bare `swift run apfel-tests` calls in `scripts/publish-release.sh:51` and `scripts/release-preflight.sh:81` are now pinned too - otherwise a release cut from this CLT-only Mac would still break on SwiftPM 6.4+.
- The policy tests in `Tests/integration/test_build_info.py` now scan `scripts/*.sh` alongside the Makefile (red on the two scripts before the fix, green after), so the gap is CI-visible.
- CHANGELOG bullet mentions the scripts.

Verified:
- `make build` (uses the pinned flag): Build complete (61.40s)
- `swift run --build-system native apfel-tests`: All 1157 tests passed
- `pytest Tests/integration/test_build_info.py test_release_pipeline.py`: 6 passed
- Model-free integration partition (`-m "not model and not serial"`) against this build: 275 passed

Not verified: the SwiftPM 6.4 failure mode itself - no 6.4 toolchain here. That half rests on the upstream report (swiftlang/swift-package-manager#10561). Body correctly says Refs #194, not Fixes, so the macOS 27 qualification issue stays open.

Approving and squash-merging per Franz's authorization for bug-fix PRs.

